PR #1, opened 2020-04-13
very tiny changes, each in a separate commit to simplify checking by humans

DaKnig:
mostly micro-optimizations

### DIFF
--- a/gbdk-lib/libc/gb/set_data.s
+++ b/gbdk-lib/libc/gb/set_data.s
@@ -106,10 +106,9 @@ _set_sprite_data::
 	ADD	HL,HL
 	ADD	HL,HL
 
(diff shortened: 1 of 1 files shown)


Zal0:
The problem I have with so many fixes is that I have no idea how to test all this

DaKnig:
what would you suggest for testing? we can make a bunch of test ROMs with something like https://github.com/ISSOtm/bgb-mooneye-tester and ROMs that test each function including edge cases. but what would we consider the "right behaviour"? the docs or the functions? the functions in some cases have bugs and dont behave according to spec.
this pull request is not so good compared to https://gist.github.com/ISSOtm/c9b0efab55108ce63bfeb775525c8c2c/ ...

PR #13, opened 2020-05-17
Add fast abs w/tests

JL2210:
The optimized version takes around 65% of the cycles that the
C version does. Included are a consistency test (which tests
every return value from this abs with the return value from a
known good one), and a benchmark (which uses the BGB macros for
accurate clock cycle counting).

As this passes both the tests I think it's safe to be used by
default.

Credit to @aaaaaa123456789 for golfing the optimized version
down to 15 bytes from my initial attempt which was 20.

### DIFF
--- a/gbdk-lib/libc/Makefile
+++ b/gbdk-lib/libc/Makefile
@@ -12,7 +12,7 @@ endif
 
 TOPDIR = ..
 
-CSRC =	abs.c atoi.c atol.c isalpha.c isdigit.c \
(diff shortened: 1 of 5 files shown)


Zal0:
Thanks
I have removed the tests, we don't need them on the library

JL2210:
You forgot to merge it into master

Zal0:
I only merge on master when doing a release

JL2210:
In that case, you should make develop the default branch so you can merge PRs automatically instead of by hand. That way, you can still keep the workflow of merging into master before a release.

PR #18, opened 2020-05-20
some library unification towards SDCC4 library

untoxa:
i re-organized some files in the library, removed duplicates, enabled assembler-written memcpy() and so on.

the sdcc-related stuff is in libc\asm\gbz80\
and gameboy-related stuff is in libc\gb\

we shouldn't mix that.


Zal0:
Merged in develop and tested

PR #19, opened 2020-05-23
interrupts fix

untoxa:
functions were perfect, except
`LDA	HL,4(SP)	; Skip return address and registers`
and
`LDA	HL,.BANKOV+2(SP); Skip return address and registers`


untoxa:
fixed multiple exceptions when run in "rom cart troubleshooting mode"

untoxa:
change the interrupt handler default behaviour

untoxa:
there is a well-known bug in DMG hardware: https://gbdev.io/pandocs/#sprite-ram-bug
we have also a bug in crt0 code, that encounters this hardware bug, that results in filling the OAM with random data on startup instead of cleanup. and the other problem - when you run your rom in"troubleshooting mode" of a BGB emu, you always receive 256 exceptions while still in initialization stub. this patch fixes both problems: OAM is cleared and no exceptions in BGB while troubleshooting the rom.

untoxa:
__BGB_PROFILER_INIT must be static

untoxa:
BGB_MESSAGE_FMT macro allows to output formatted strings

Zal0:
> 
> 
> there is a well-known bug in DMG hardware: https://gbdev.io/pandocs/#sprite-ram-bug
> we have also a bug in crt0 code, that encounters this hardware bug, that results in filling the OAM with random data on startup instead of OAM cleanup. and the other problem - when you run your rom in"troubleshooting mode" of a BGB emu, you always receive 256 exceptions while still in initialization stub. this patch fixes both problems: OAM is cleared and no exceptions in BGB while troubleshooting the rom.

I am still getting those 255 warnings, did you test this LOL?

Zal0:
All changes merged

Issue #29, opened 2020-06-14
lcc determines it's directory wrongly


basxto:
Calling lcc as `gbdk-2020/build/gbdk/bin//lcc` does not work, but there shouldn't be a difference to `gbdk-2020/build/gbdk/bin/lcc`

basxto:
https://github.com/Zal0/gbdk-2020/blob/v3.2/gbdk-support/lcc/gb.c#L281 is causing this

Zal0:
why do you want to call lcc that way?

basxto:
That’s just something that tends to happen when you concatenate variables to paths, you often add extra `/` for safety reasons when variables can be changed by the user

untoxa:
fixed in 4.0 release

PR #33, opened 2020-06-15
Unify bin directories and improve lcc

basxto:
- merge `bin` and `bin/SDCC/bin`
- support `//`, `/./`, `/../` with `lcc`
- add `--sdccbin=` to lcc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ ASDIR = $(TOPDIR)/as
 
 # Base setup
 # Extension to add to executables
-EXEEXTENSION = 
+EXEEXTENSION =
 # Host operating system identifier.  The one below should work for
 # most systems.
 HOSTOS = ppc-unknown-linux2.2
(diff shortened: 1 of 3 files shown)


Zal0:
Why do you want to unify sdcc/bin into bin?
Updating SDCC it's easier if it's on a separate folder

basxto:
1. `bin` directories are unified so it is compatible with UNIX file structure `${prefix}/bin` `${prefix}/lib`, which can be `/bin`, `/usr/bin` etc.
2. Fixes #29
3. `--sdccbin=` is for the case sdcc and gdbk are in different folders, which happens when you build your own portable gbdk and use installed sdcc or if one wants to use it with a newer sdcc version in case it dies like original gbdk.

With just 3. it would be neccessary to write a wrapper script when you want to put lcc in `/usr/bin`. Without 1. and 3. it would be neccessary to write a wrapper script and to create symbolic links to all sdcc executables.

Why is it simpler, don’t you just overwrite the files?

Zal0:
Thanks a lot. I'll test it on Windows and MacOS and merge it
Does this fix Linux packaging issues?

basxto:
Nearly, I still have to modify `bindir` in `gb.c`.
And it looks like `--help` is broken, but that was already the case in `v3.2`.

basxto:
I wrote an  arch linux package based on that.

basxto:
With that update it can now be built without manipulating any files.

basxto:
Windows should use the old directory again.
I hope it’s mergable this way.

untoxa:
fixed other way in v4.0 release

PR #35, opened 2020-06-21
LCD interrupt must NOT wait for the compatible LCD mode

untoxa:
LCD interrupt must NOT wait for the compatible LCD mode, because that excludes the possibility of making effects, that require every HBlank interrupt. wobble.c is an example of such an effect written in pure C. Compile wobble.c with and without a patch to see the difference.


untoxa:
nowait_int_handler() does not help here, because the resulting construction too slow for an HBlank, which must be less than 114 cycles.

Zal0:
so... should we merge any of this?

untoxa:
yes, that's why we need to change the default behaviour for an LCD handler (only that handler). to make possible some minimal reasonable interrupt. i think that interrupt handler chain for LCD interrupt is useless, maybe we must make only one interrupt handler for LCD possible ( but leaving the compatible API: add_LCD(), remove_LCD() ) - that saves the cycles and allows to write a bit more complex handlers within that very short time intervals? iterating through the vector "eats" too much.

untoxa:
I moved the LCD interrupt handler out of crt0. The changes are compatible and transparent.

untoxa:
".globl    .add_LCD" must be commented out, since we can have custom hardcoded handlers now.

Zal0:
This has been merged to develop now

PR #38, opened 2020-07-02
Fix MBC5 for 32K ROMs

basxto:
This assumes main being in bank 1 instead of 0
Bank 0 is always loaded anyways
- should fix the MBC write exception in BGB
- allows 32K ROMs on MBC5 without manually switching to “bank 1” at the start of `main()`
- allows 32K ROMs on MBC5 where main moved into “bank 1”, even though I have my doubts this could happen

### DIFF
--- a/gbdk-lib/libc/gb/crt0.s
+++ b/gbdk-lib/libc/gb/crt0.s
@@ -279,7 +279,7 @@ banked_call::			; Performs a long call.
 	.if __RGBDS__
 	.dw	BANK(_main)
 	.else
-	.dw	0
+	.dw	#1
 	.endif
 _exit::	
 99$:


basxto:
This apparently breaks farfixer.
I guess 32K ROMs have to `SWITCH_ROM_MBC1(1);` when main starts. This works since toxa moved `banked_call`.
But the ROM might break if main returns and is >16KB, since `banked_ret` will switch back to 0.

bbbbbr:
I'm going to keep an eye on this for when I eventually migrate to the new GBDK. Seems like I may have to watch out for it.

I have a project that's currently using a 32K ROM in non-MBC unbanked mode. _CODE is a full 32K instead of 16K, and  _CODE_1 is *not* used. (-Wl-yt0 -Wl-yo0).

If there are some special cases here it would also be a good idea to have them formally documented.

basxto:
If you don’t need banking, MBC1 could do the job.
If you need banking, you will bank your images, music etc, which is currently part of your _CODE

bbbbbr:
I was targeting a non-mapper cartridge such as the one below. Does that mean it's no longer an option in the new GBDK (as of now)?

https://catskullelectronics.com/32kcart

```These carts are dead simple, only capable of handling roms up to 32k in size. There is no mapper, no SRAM, no battery, no RTC. Just a rom. ...```

basxto:
No, there is no problem without any MBC and with MBCs <5

bbbbbr:
Thanks!

PR #39, opened 2020-07-06
optimized joypad routines; register usage hints; RST38 crash handler

untoxa:
shorter and more optimal code; hints in gb.h allow better register allocation by SDCC - we might set that for all frequent asm functions.


untoxa:
that register hints help SDCC generate more optimal code

untoxa:
crash handler support, handler implementation and usage example

untoxa:
more optimal memcpy()

untoxa:
fast uint2bcd(), similar might be used with printf() to eliminate very slow div/mod.

PR #43, opened 2020-07-12
Rewrote _strcmp to fix bug optimize it slightly.

nitro2k01:
Fixed bug where first character in string is ignored. Optimized the code slightly.


untoxa:
asm_strings.s was splitted, patch should be applied to _strcmp.s

untoxa:
merged with another file here: https://github.com/Zal0/gbdk-2020/commit/90f05d40d63e4ce39f49f923ebcb671ebbfaa666

PR #48, opened 2020-08-30
various library enhancements

untoxa:
…provement, smaller code size.

.OAM name changed to _shadow_OAM - must pass that into the linker through the commandline: -g _shadow_OAM=0xC000


untoxa:
codegeneration for those inline functions is reasonable, definitely they worth to be used that way.

untoxa:
new trampoline for current builds of SDCC

untoxa:
multiple library objects are in one lib now

untoxa:
12 times faster itoa()

PR #57, opened 2020-10-05
Fix -S flag in LCC

### LINKED ISSUE #52: lcc -S doesn't give assembly
Store in `abs.c`:
```
int abs(int a)
{
    return a < 0 ? -a : a;
}
```

run `lcc -S test.c`; nothing new in the directory
run `lcc -S test.c -o test.asm`; no test.asm created
run `lcc -S test.c -o -`; assembly not output to stdout

running GBDK v4.0 on Linux

JL2210:
Fixes #52 (introduced in commit https://github.com/Zal0/gbdk-2020/commit/4928c8884bd2cc05130bce816f203473c7cf925a)

While I'm at it, fix some other things:
* use header files to get function prototypes rather than defining them ourselves
* Include stdlib.h in gb.c to get the definition for exit
* prevent several possible overflows
* use proper types for PIDs

### DIFF
--- a/gbdk-support/lcc/gb.c
+++ b/gbdk-support/lcc/gb.c
@@ -2,6 +2,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 #include <assert.h>
 #include <ctype.h>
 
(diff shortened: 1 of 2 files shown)


Zal0:
Thanks!

JL2210:
@Zal0 This is still an issue in 4.0 release. May I suggest making 4.1?

Zal0:
Yes
There are still some issues with people testing some stuff, when develop becomes a bit more stable we should release 4.1

PR #63, opened 2020-10-21
Update GBDK Docs

bbbbbr:
**NOTE : This is more of a Request For Comment / Discussion than a PR.**

The existing, old GBDK docs would really benefit from being updated. Some parts are not well or fully documented, others are out of date versus GBDK 2020 and the appearance is generally dated.
http://gbdk.sourceforge.net/doc/html/index.html
http://gbdk.sourceforge.net/doc/html/book01.html
http://gbdk.sourceforge.net/guidelines.html

A lot of the existing documentation was generated with Doxygen from the Header files. Doxygen appears to still be underactive development and is easy to use. So, it seems worth exploring whether that could continue to be a workable solution.

Here is a proof of concept to see what it might look like, both in how it affects source code and what it could look like.
- Updated and filled out many comments for functions and macros in gb.h. (See content of PR)
- Replaced the default Doxygen theme (looks ~circa 2005-2010) with one that roughly matches Github's Markdown style, and the site more generally after a couple improvements.

For example:
https://bbbbbr.github.io/gbdk-2020-docs/api/gb_8h.html#a65971730b7a58df99fc43be0162226ac

From this repository:
https://github.com/bbbbbr/gbdk-2020-docs

I'm interested in what folks think of this, or if they have other ideas or approaches.

Some questions to consider:
- Use of markdown files for non-code generated docs, or try to use Doxygen to generate non-API content as well (coding guidelines, toolchain info, etc)
- The additions to gb.h greatly increased the file length (by about 400 lines!). Maybe that is ok. Or maybe not. It might be possible to locate some of that in .c files instead. 
- Have Docs as a separate repository, or a subset of the main one?

Perhaps more later, but this is enough for a start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ link/Debug/
 link/.vs/
 far_fixer/Debug/
 far_fixer/.vs/
+
+# Doc generation files
+gbdk-lib/doc/
--- a/gbdk-lib/include/gb/gb.h
+++ b/gbdk-lib/include/gb/gb.h
@@ -35,48 +35,62 @@
 
 /** Screen modes.
     Normally used by internal functions only.
+    @see mode()
  */
 #define	M_DRAWING    0x01U
 #define	M_TEXT_OUT   0x02U
 #define	M_TEXT_INOUT 0x03U
-/** Set this in addition to the others to disable scrolling 
-    If scrolling is disabled, the cursor returns to (0,0) */
+/** Set this in addition to the others to disable scrolling
+
+    If scrolling is disabled, the cursor returns to (0,0)
+    @see mode()
+*/
 #define M_NO_SCROLL  0x04U
-/** Set this to disable \n interpretation */
+/** Set this to disable interpretation
+    @see mode()
+*/
 #define M_NO_INTERP  0x08U
 
-/** If this is set, sprite colours come from OBJ1PAL.  Else
-    they come from OBJ0PAL.
+/** If this is set, sprite colours come from OBJ1PAL. Else
+    they come from OBJ0PAL
+    @see set_sprite_prop().
 */
 #define S_PALETTE    0x10U
 /** If set the sprite will be flipped horizontally.
+    @see set_sprite_prop()
  */
 #define S_FLIPX      0x20U
 /** If set the sprite will be flipped vertically.
+    @see set_sprite_prop()
  */
 #define S_FLIPY      0x40U
 /** If this bit is clear, then the sprite will be displayed
-    ontop of the background and window.
+    on top of the background and window.
+    @see set_sprite_prop()
 */
 #define S_PRIORITY   0x80U
 
 /* Interrupt flags */
-/** Vertical blank interrupt.
-    Occurs at the start of the vertical blank.  During this
-    period the video ram may be freely accessed.
+/** VBlank Interrupt occurs at the start of the vertical blank.
+
+    During this period the video ram may be freely accessed.
+    @see set_interrupts(), @see add_VBL
  */
 #define VBL_IFLAG    0x01U
-/** Interrupt when triggered by the STAT register.
-    See the Pan doc.
+/** LCD Interrupt when triggered by the STAT register.
+    @see set_interrupts(), @see add_LCD
 */
 #define LCD_IFLAG    0x02U
-/** Interrupt when the timer TIMA overflows.
+/** Timer Interrupt when the timer @ref TIMA_REG overflows.
+    @see set_interrupts(), @see add_TIM
  */
 #define TIM_IFLAG    0x04U
-/** Occurs when the serial transfer has completed.
+/** Serial Link Interrupt occurs when the serial transfer has completed.
+    @see set_interrupts(), @see add_SIO
  */
 #define SIO_IFLAG    0x08U
-/** Occurs on a transition of the keypad.
+/** Joypad Interrupt occurs on a transition of the keypad.
+    @see set_interrupts(), @see add_JOY
  */
 #define JOY_IFLAG    0x10U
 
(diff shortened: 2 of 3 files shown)


untoxa:
that's awesome! it it possible to generate and keep up-to-date github wiki with doxygen too?

bbbbbr:
By "github wiki" do you specifically mean the built-in github wiki functionality (https://guides.github.com/features/wikis/), or the example docs website linked above? (https://bbbbbr.github.io/gbdk-2020-docs/api/gb_8h.html)

If you mean the built-in github wiki functionality, it isn't able to interface with that as far as I know.

If you mean the example docs website linked above, then as long as comments in the source code are kept current it is possible to generate updated documentation at any time. The example uses the github Pages (https://pages.github.com/) feature to read from the github-themed doxygen output. Regenerating the docs could be part of the main build process, or a secondary optional step.

Where the output is stored is part of the discussion: 
- In the same repository as the main source? 
- An external repository such as in the example above? 
- As sub-module of the main one?

untoxa:
i meant the built-in github wiki. like: https://github.com/Zal0/ZGB/wiki/Animations
but github.io also goes, i just registered https://github.com/gbdk-2020, so we may have something like gbdk-2020.github.io/docs/...

Zal0:
Personally  I don't like github wikis
The link is a bit hidden and most people are not aware of its existance
Also they are in a different repo already so keeping it synchronized is complicated. In ZGB I have plans to fully remove it and instead have it inside the docs folder. This way the docs will be always synchronized with the current branch

I think we should do the same with gbdk. Just creating a docs folder and put it there. Then in the readme.md we can directly add a link to it

untoxa:
can this be organized like:
```
git pull
make docs
cd docs
git push
```
and that automatically updates `docs` repo and that magically appear on the website here: https://gbdk-2020.github.io/docs/ ?
so we remove generating of docs out of the main repo, only handle comments carefully, and all the docs generating stuff will be in https://github.io/gbdk-2020/gbdk-2020-docgen/ repo?

that `gbdk-2020-docgen` should have two repos: `Zal0/gbdk-2020` and `gbdk-2020/docs` as submodules?

Zal0:
Submodules are a pain
Why not having everything in the same repo? That way docs can be updated when making a feature and that update will be automatically merged (together with the docs) when merging into master.
We could also force people to fully document their features before merging into pr

untoxa:
that means all that doxygen stuff must be in the main repo.

"force people to document features" mean "write proper comments inside the headers" - there is no problem, just don't accept pr's that do not meet requirements. that is a totally different thing with "building the docs from the sources and make them accessible online".

online docs with simple obvious url is a good feature. folder with html or pdf files inside a repo is good, but not the same as @bbbbbr suggests.

bbbbbr:
I am somewhat  leaning towards Zal0's perspective of a single repo.

__Pros:__ Everything is together in one simple package
__Cons:__ Repo is a little larger, occasionally there will be more commit traffic due doc updates.

__Example:__ Here is a test version with the docs and doxygen config stored under a /docs folder in main repo. Github Pages root is set to /docs/. I put the generated docs under /docs/api, but they could just as easily be in the /docs root instead. This example just leaves room for other general content.
https://bbbbbr.github.io/gbdk-2020/
https://github.com/bbbbbr/gbdk-2020/tree/develop_docs_in_main_repo

Git sub-modules: I have mixed feelings about them, the implementation in git seems kind of awkward at times. *shrug*


__Contributors & Docs__ I agree with toxa that *generated doc updates* should not be mandatory when submitting commits- as long as the commits contain the required and properly formatted header file updates. If contributors want to include updates that should probably be ok too.

__Releases & Updating Docs__: It is probably sufficient to generate updated docs once per release, then push those to whatever branch is attached to the website on Github Pages. That way the main displayed docs match whatever the current official release it.

Docgen setup/tools + docs output: I would encourage keeping these together and not splitting them into separate repos. The doxygen config and theme are only a couple files.

__Github Wikis__: I don't see much advantage to using them versus Github Pages. There appears to be less flexibility in how they are generated,  stored and served. (could be wrong about that)

__Github Pages__: You can point the base url at either the root of a repo, or a 1st level sub-folder (such as docs) so that it's served as the main page. That makes it possible to give an obvious / default URL to content that is in /docs/. Pages can serve both HTML and markdown formatted content, so it can even be a mix of Doxygen output + markdown pages if that is preferred, instead of just using Doxygen to make all the docs.

"write proper comments inside the headers" - Providing an example demonstrating the expected standard should be easy. Many projects have PR submission guidelines, this could be part of them.

__The gbdk-2020 account__: Regardless of how it's used, it seems useful to have a somewhat more official looking project home.

untoxa:
i am about to merge this commit. but the question about gbdk-2020 account is still open. if we make docs in the main repo, how do we make them mirrored in gbdk-2020.github.io/docs ?

untoxa (approved):

untoxa (approved):

bbbbbr:
> i am about to merge this pr. but the question about gbdk-2020 account is still open. if we make docs in the main repo, how do we make them mirrored in gbdk-2020.github.io/docs ?

gbdk-2020/docs would have to basically be a cloned repo of the zal0/gbdk-2020. 

A different approach to a named URL would be just getting a domain name, such as "gbdk2020.io". Then it (and subdomains) can be pointed at any underlying github account + repo. So it could be gbdk2020.io/docs. Similar to what is done with : http://gbdev.io , http://hh.gbdev.io etc. That has more long term flexibility since it can be changed to point to different repos (or hosting providers) in the future if needed. It does of course require a little $ and remembering to not let the domain expire in the future.

untoxa:
@Zal0 should decide that

bbbbbr:
That makes sense. 

I will have more commits with updated docs in the future (I'll try to consolidate them to a few commits)

I can also open a PR for the __develop_docs_in_main_repo__ branch if that becomes the definite course of action.

PR #65, opened 2020-10-24
Expose the default Serial Link ISR so that it can easily be added and removed

bbbbbr:
This just exposes serial_IO so that it can be used with remove_SIO() and add_SIO() in case you want to replace the default Serial Link handler. serial_IO is the ISR for the Serial Link that gets installed by default if any of the standard API for SIO is used.

@untoxa Mentioned that if you don't use any of the standard API for SIO and write your own ISR then the default handler doesn't get installed automatically. However, I do think it's useful to still have the ability to add and remove it.

### DIFF
--- a/gbdk-lib/include/gb/gb.h
+++ b/gbdk-lib/include/gb/gb.h
@@ -39,7 +39,7 @@
 #define	M_DRAWING    0x01U
 #define	M_TEXT_OUT   0x02U
 #define	M_TEXT_INOUT 0x03U
-/** Set this in addition to the others to disable scrolling 
+/** Set this in addition to the others to disable scrolling
     If scrolling is disabled, the cursor returns to (0,0) */
 #define M_NO_SCROLL  0x04U
 /** Set this to disable \n interpretation */
@@ -145,7 +145,7 @@ void add_LCD(int_handler h) NONBANKED;
     changes from $FF to $00.
 
     @see add_VBL
-*/    
+*/
 void add_TIM(int_handler h) NONBANKED;
 
 /** Adds a serial transmit complete interrupt handler.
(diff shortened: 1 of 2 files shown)


untoxa:
I want to make isr chain handling api oblolete. In the new code you should not do that.

untoxa:
Renaming the “.” symbol to “_” symbol is not a good idea because this might break backward compatibility with 3-rd party asm code. In such cases it is better to introduce aliases. We also have default VBlank handler, that is invisible in C, and in theory, one might need to remove it for some reason (that will break half of the library, though).

untoxa:
sorry, @bbbbbr, i really think this is useless. you are adding a handler that is only needed for removing default functionality that is ok: see comm.c example. if you are skilled enough to write your own default interrupt handler for serial io, then you just use new better technique. that additional symbol in gb.h will only confuse people.

bbbbbr:
Good point about the symbol naming and an alias.

Does your plan for ISRs still include a way to install/remove them at runtime and not just link time? That is, a way to support multiple different ISRs which can be swapped out, even if only one is installed at a time. It is probably not common to do that with SIO/JOY/TIM, but seems more common to swap them when used for effects/etc with VBL and LCD.

An example: https://github.com/mills32/CUTE_DEMO/search?q=remove_VBL

untoxa:
this is faster:
```
volatile char music_on = 0, something_else = 0;
void my_timer_isr() __critical _interrupt {
  if (music_on) play_music();
  if (something_else) do_something_else();
}
void play_music() {
...
}
void do_something_else() {
}
```
or in that particulare case from mills demo you might have only one pointer at the moment:
```
handler_t handler = 0;
void my_timer_isr() __critical _interrupt {
  if (handler) handler();
}
void play_music() {
...
}
void do_something_else() {
}
void main() {
  __critical { handler = play_music };
  while(1);
}
```
you rarely need chains.

bbbbbr:
A couple more questions - 

Currently TIM / JOY / SIO pass through .int:: and so now benefit from the mode-exit protection by default (unless turned off with nowait...). In what you describe above, would that still apply or would the expectation be when writing an ISR that if mode-exit protection is wanted it should be written at the end of the ISR?

And, for clarity, is your plan to remove the add_xxx() and remove_xxx() ISR functions, or just discourage use and not maintain them (but leave them operable)?

untoxa:
of course i'm not going to remove them. and i'm going to fix that remove function too and keed everything operable.

i think `while (STAT_REG & 0x02);` should work. i'm thinking about more efficient ways, like custom exit code, maybe.

untoxa:
https://github.com/Zal0/gbdk-2020/commit/3410b6183346136382d4faee83e1c1c8a7d978cb

PR #76, opened 2020-11-05
Move Example Projects into separate folders and give them individual makefiles

bbbbbr:
Per request by @untoxa this PR moves each example project to it's own subdirectory.

They all get individual Makefiles, and the Makefile in the example root will try to build all the example subdirectories.

Tested and working on Linux & Windows :heavy_check_mark: 

Slightly renamed some samples to clarify their purpose and group them, such as sgb_* examples, lcd_isr_wobble, etc.

Applied some minor build fixes to a couple samples that had accumulated small build errors from API changes. such as:
* /gb-dtmf.c : local init_bkg conflict with new lib function, renamed to init_background()
* Fixing include path in space.s

There were a couple problems that came up with the samples when test-building them. I might look into some of these.
 * rand.c (uninitialized wram exception)
 * sgb_border.c (uninitialized wram exception)
 * rpn.c (keyboard location looks incorrect vs expected)
 * sgb pong (super gameboy detection doesn't seem to work in bgb?)
 * The fprpn.c sample has building turned off since it has build problems related to needing floating point div/mul/etc. Looks like this is not supported in gbdk right now? Maybe just remove this sample?
 

Suggestions for the future:
 * To help catch build errors with examples consider also building them while doing the main gbdk build. 
 (Doing that  would require changing their Makefiles to use an environment var instead of the current relative path to lcc.exe)

* Remove Makefile.rgbds since it is apparently not used and does not work.

* There are some very simple examples here that could be useful additions, either by linking to the repo from the docs or incorporating them if the author is open to that. Not sure if they are compatible with GBDK 2020 yet. https://github.com/mrombout/gbdk_playground


untoxa:
almost perfect. farptr, not farprt :)))

bbbbbr:
Thanks! - Fix in #75

Issue #72, opened 2020-11-03
bank overflow checking in LCC


untoxa:
that should be done in SDCC's makebin tool, but we might wait for a decade for that, so, it is nice to implement that ourselves instead, that is easy: https://gbdev.gg8.se/forums/viewtopic.php?pid=4540#p4540

bbbbbr:
A first pass at a small tool for checking IHX overlaps. It builds for Linux and Windows (mingw).
https://github.com/bbbbbr/ihx_check

- It does some merging of adjacent data blocks to reduce the number of comparisons, since in the ihx file they typically 32 bytes at the largest. That can be changed if we want to test all of them individually.

- By default it just prints warnings without changing an error code, but it can be instructed to treat the warnings as errors with the `-e` switch after the .ihx filename.

Some example output with a couple banks overflowing into others by various amounts.
```$ ihx_check.exe banks.ihx
Warning: Multiple write of 16384 bytes at 0x8000 -> 0xbfff. Bank overflow?  (8000 - bfff and 8000 - bfff)
Warning: Multiple write of     1 bytes at 0xc000 -> 0xc000. Bank overflow?  (c000 - c01e and c000 - c000)
Warning: Multiple write of    28 bytes at 0xc000 -> 0xc01b. Bank overflow?  (c000 - c01e and c000 - c01b)
Warning: Multiple write of     1 bytes at 0xc000 -> 0xc000. Bank overflow?  (c000 - c000 and c000 - c01b)
```

untoxa:
"Bank overflow?" -- you age guessing? :)
```
.area area1
.org 0x500
.db 0, 1, 2

.area area2
.org 0x501
.db 3
```
nothing overflows here. just output addresses and values like the old tool did.

bbbbbr:
@Zal0 Any preferences on this tool?

The "Bank overflow?" text was meant as a hint to the user about what might be going wrong (a question that occurs) and the title of this ticket does happen to be "bank overflow checking". :) No problem though, I can remove the hinting.

I could also change to printing "addresses and values", but a couple points in favor of summarizing the overlap instead:
* Since the address overlap can run into the 100-1000s of bytes, summarizing provides a more compact warning that doesn't scroll for multiple pages.
* Reporting the overlap size in bytes also makes it easier to figure out how much size needs to be reduced in cases of overflow (such as https://gbdev.gg8.se/forums/viewtopic.php?pid=4540)
[...]

PR #85, opened 2020-11-12
Add multiple write check for IHX files (ihxcheck) to build process

bbbbbr:
* Add ihx multiple write tool "ihxcheck" tool to gbdk-support
* Build ihxcheck with main Makefile
* Call ihxcheck from Lcc after ihx file has been generated with linker
* Allow passing an argument to ihxcheck via lcc (-Wi-e) to error out instead of just warn on multiple writes


bbbbbr:
BTW, this is a candidate for #72

untoxa:
need a simple way to switch this tool off with a commandline argument

bbbbbr:
That could be either a primary Lcc switch or a pass-through switch to ihxcheck which causes it to not process and exit immediately. Any preference?

untoxa:
LCC switch

bbbbbr:
Switch name is easy to change if K isn't desired.

PR #87, opened 2020-11-14
Retrieve system identifier from "uname"

jserv:
Both host and target operating system informations were hardcoded. This
patch attempts to retrieve them via "uname -s".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,10 @@ DOXYGEN_VER_HAS = $(shell doxygen -v)
 # Base setup
 # Extension to add to executables
 EXEEXTENSION =
-# Host operating system identifier.  The one below should work for
-# most systems.
-HOSTOS = ppc-unknown-linux2.2
+# Host operating system identifier.
+HOSTOS = $(shell uname -s)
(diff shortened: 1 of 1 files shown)


Zal0:
Can someone confirm if this works on the 3 OS: Windows, Macos and Linux ? (specially windows)

DaSalba:
Was going to check it on Linux, but seems it has already been merged, right?

bbbbbr:
It does seem to be merged already.

Note: the uname output it looks like it would only be for Linux & MacOS.  For Windows MinGW builds TARGETOS gets overwritten at the location below (and HOSTOS is not used for anything except setting TARGETOS).
```
cross-linux-mingw32-build:
	$(MAKE) TARGETOS=i586-mingw32msvc \
```
https://github.com/Zal0/gbdk-2020/blob/6723f7f531c9cf7b553f32166484a5b0c8f3737e/Makefile#L85

Also this for the expected output:
https://stackoverflow.com/questions/3466166/how-to-check-if-running-in-cygwin-mac-or-linux

PR #88, opened 2020-11-14
Improve the rules to install SDCC executables

jserv:
This patch specifies the predefined list for SDCC executable files for
GBDK installation. For x86_64-w64-mingw32 target, the extra DLL files
are copied as the necessary runtime.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,27 +171,23 @@ gbdk-lib-examples-makefile:
 	$(MAKE) -C $(BUILDDIR)/examples/gb make.bat
 	unix2dos $(BUILDDIR)/examples/gb/make.bat
 
-#SDDC copy
+# Copy SDDC executable files
+SDCC_BINS = as2gbmap makebin packihx sdar sdasgb sdcc sdcdb sdcdb sdcpp sdldgb sdnm sdobjcopy sdranlib sz80
+ifeq ($(OS),Windows_NT)
+MINGW64_RUNTIME = \
+	libgcc_s_seh-1.dll \
+	libgcc_s_sjlj-1.dll \
+	libstdc++-6.dll \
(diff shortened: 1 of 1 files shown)


untoxa:
for windows append ".exe" to every item in the list. it should also copy dll's, if they exist: 
libgcc_s_seh-1.dll 
libgcc_s_sjlj-1.dll 
libstdc++-6.dll 
libwinpthread-1.dll 
readline5.dll 
they might not, if compiled statically - that's not an error.

jserv:
> for windows append ".exe" to every item in the list. it should also copy dll's, if they exist:
> libgcc_s_seh-1.dll
> libgcc_s_sjlj-1.dll
> libstdc++-6.dll
> libwinpthread-1.dll
> readline5.dll
> they might not, if compiled statically - that's not an error.

I extracted the [SDCC snapshot files](http://sdcc.sourceforge.net/snap.php) for `i586-mingw32msvc`.  There is only one `readline5.dll` in `sdcc/bin` directory while `x86_64-w64-mingw32` contains the above DLL files. Shall I support `i586-mingw32msvc` as well?

untoxa:
we build only Win64 binaries and use Win64 version of SDCC

jserv:
Thank @untoxa for clarification. I have force-pushed.

untoxa:
i suggest to include z80 target, because we have infrastructure for that in makefiles. maybe, someday, we extend library to support SMS. the list of tools:
```
makebin.exe
packihx.exe
sdar.exe
sdasgb.exe
sdasz80.exe
sdcc.exe
sdcdb.exe
sdcpp.exe
sdldgb.exe
sdldz80.exe
sdnm.exe
sdobjcopy.exe
sdranlib.exe
sz80.exe
```

PR #95, opened 2020-11-16
Examples: Add bgb debug example

bbbbbr:
It could be a useful example to include since it's now part of the API.

Some of the code is from the GB forums post on the topic (toxa, zal0).


untoxa:
wrong and not needed gbdk-lib/examples/gb/bgb_debug/make.bat here

jserv (changes requested):
File "make.bat" should not be checked in because it was generated automatically by Makefile. Maybe you can submit another pull request which takes ".gitignore" to exclude such generated files.

bbbbbr:
Bleh, didn't mean to include that.  Removed.

> Maybe you can submit another pull request which takes ".gitignore" to exclude such generated files.

Good idea

PR #97, opened 2020-11-18
Fix common realloc mistake

jserv:
The assigned variable would be nulled but not freed upon failure.


Zal0:
Not sure about the chances of realloc failing in this scenario and there is an assert too... but ok

PR #104, opened 2020-12-05
BankPack: Automatic bank assignment / packing

bbbbbr:
OK, as discussed so far here is a first pass at automatic bank packing. I won't be surprised if changes are required.
@Zal0 @untoxa 

__It would be good to have testing from some other people__

* Has been built and tested on both Windows x64 and Linux x64, seems ok so far
* LCC: --autobank option turns on bank packing
* LCC: if present then -ytN will be automatically copied into the bankpack argument list to set the mbc type. That will  automatically update the maximum available bank based on the determined MBC (it can be overridden).
* An example project has been added: banks_autobank\

__Features__:
* For a given file set the ROM bank to 255 in order for it to get auto-banked (when enabled). For example: `#pragma bank 255 at the top of the file`. Also: `const void __at(255) __bank_<const name of your choice>;` (such as filename without extension).
* If the bank is set to something else (such as `4`) then bankpack will _only_ count the size used for that bank and  leave the file untouched. This means it is possible to intermingle both fixed and auto-banked files in a project if needed.
* S b_<symbol> Def0000FF gets updated when a matching non-banked version of the symbol is found
* ___bank_<symbol> Def0000FF gets updated. Additional prefixes other than `___bank_` can be added using -sym=
* There are more options, as shown below in the help output

__Note__:
* On occasion it may require a `make clean; make` in order to reset previously packed object files and re-pack everything (once a bank is changed for an object file from 255 to X it will remain that way until being recompiled which resets the bank)
* Only one assigned bank allowed per-file

```
$ ./bankpack  -h
bankalloc [options] objfile1 objfile2 etc
Use: Read .o files and auto-assign areas with bank=255.
     Typically called by Lcc compiler driver before linker.

Options
-h           : Show this help
-yt<hexbyte> : Set MBC type per ROM byte 149 in Hex (see pandocs)
-mbc=N       : Similar to -yt, but sets MBC type directly to N instead
               of by intepreting ROM byte 149
               mbc1 will exclude banks {0x20,0x40,0x60} max=127, 
               mbc2 max=15, mbc3 max=127, mbc5 max=255 (not 511!) 
-min=N       : Min assigned ROM bank is N (default 1)
-max=N       : Max assigned ROM bank is N, error if exceeded
-ext=<.ext>  : Write files out with <.ext> instead of source extension
-path=<path> : Write files out to <path> (<path> *MUST* already exist)
-sym=<prefix>: Add symbols starting with <prefix> to match + update list.
               Default entry is "___bank_" (see below)
-v           : Verbose output, show assignments

Example: "bankpack -ext=.rel -path=some/newpath/ file1.o file2.o"
Unless -ext or -path specify otherwise, input files are overwritten.

Default MBC type: no MBC

The following will have FF and 255 replaced with the assigned bank:
A _CODE_255 size <size> flags <flags> addr <address>
S b_<function name> Def0000FF
S ___bank_<const name> Def0000FF
    (Above can be made by: const void __at(255) __bank_<const name>;
```


untoxa (approved):

untoxa:
Need possibility for returning number of banks occupied back into lcc, to pass that into makebin.

PR #105, opened 2020-12-09
BankPack: Calculate cart size (-cartsize)

bbbbbr:
@untoxa 

Adds auto-calculation of cartsize. It is rendered in stdout as "autocartsize:N", warn about fixed banks outside specified limits, fixes a recent typo in a lcc comment

(I could be wrong about this...)
For LCC it doesn't appear straightforward to capture STDOUT from the spawned bankpack child process, particularly because commands get executed in a different way between Windows and Linux.
* On Windows LCC uses _spawnvp() which is part of the API. I didn't see obvious support for capturing STDOUT from the child.
* On Linux LCC uses a custom function to emulate the behavior of _spawnvp(). With that function it may be do-able using dup2() and a pipe.

### DIFF
--- a/gbdk-support/bankpack/bankpack.c
+++ b/gbdk-support/bankpack/bankpack.c
@@ -12,6 +12,7 @@
 #include "files.h"
 
 bool g_option_verbose = false;
+bool g_option_cartsize = false;
 
 static void display_help(void);
 static int handle_args(int argc, char * argv[]);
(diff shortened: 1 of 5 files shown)


untoxa (approved):

untoxa:
lets not touch lcc at the moment

bbbbbr:
If sdcc will accept a patch, modifying makebin with an auto-size option might be relatively easy.

PR #111, opened 2020-12-23
Readme: Add discord server link

bbbbbr:
Wasn't sure whether or not to mention ZGB in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@ You don't need the sources, unless you wish to compile GBDK-2020 yourself. Downl
 
 ## [Online documentation](https://zal0.github.io/gbdk-2020/)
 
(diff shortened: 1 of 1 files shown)


Zal0:
Hmmm
Have we removed the link to gbdev? I think both links should  appear

untoxa:
no, nothing is removed by that commit.

Zal0:
Ok... Right now it's a bot confusing because that link is up there but there is a link to the other server in the docs section

bbbbbr:
I'll submit another PR that brings both sections together. Same for the docs, since those have two entries as well.

bbbbbr:
Fixes in #112

PR #125, opened 2021-01-29
Change www root to serve main repo Readme.md

bbbbbr:
@Zal0 @untoxa 
Changes to allow the Readme.md in the repo root to be served as the default landing page for the github.io site that is served.

The current default www page is very sparse and doesn't have much info. It seems somewhat unnecessary. Especially since the general docs pages that I thought might get added to it I've instead merged into the doxygen content.

And, the main Readme has a lot of the useful info that would be equally useful served by www. This isn't a necessary change, but would be an improvement.

Compare: https://zal0.github.io/gbdk-2020/
Preview of changes: https://bbbbbr.github.io/gbdk-2020/

**If this change is made**: Once merged into master --> the www root should be changed in settings from `/docs/` to just `/`


bbbbbr:
@Zal0 @untoxa

Don't forget to update the www root from `/docs/` to just `/`. I guess the repo serves devel by default, not master, my mistake. So the readme link for docs is broken right now.

PR #136, opened 2021-02-14
Allow to specify directory for executables

basxto:
This allows to give a different directory for executables like `/usr/bin/`, so that the binaries and lib will end up in different folders.
This is very important for packaging gbdk-2020, since binaries will end up `/usr/bin` or a similar directory. (The ugly alternative would be to place symlinks there to lcc and symlinks in gbdk-2020's bin directory to sdcc.)

Those two commits are cherrypicked from #33 

The other commits are just there because it's based on tag 4.0.2 and develop out of sync with master.


bbbbbr:
The change looks small and isolated, yet lcc path handling is a little complex and nested (for me at least to grasp at a glance) so it would be good to make sure there are no side effects. Pardon all the questions. :)

Can you describe the layout of different binaries a little more, as an example to illustrate and test with?

Where would lcc.exe be?
Where would sdcc binaries be?
Where would ihxcheck, bankpack, etc be?
Where would inc and lib be? 

How does it interact with set_gbdk_dir()? 
Has it been tested on Windows? (Maybe not an issue if this is expected to be Linux only, but then should be noted)

And, could any of these be used as an alternative to this change?
GBDKDIR env var
--gbdklibdir=
--gbdkincludedir=

Or, would adding an override for sdccdir be more flexible instead of compiling the alternative directory into the executable? 
```
else if ((tail = starts_with(arg, "--sdccdir="))) {
    setTokenVal("sdccdir", tail);
    return 1;
}```

basxto:
> Where would lcc.exe be?
It's intended to also be in `$(BINDIR)`, but it does not matter, since it could be everywhere. This is about the hard coded paths, where lcc expects the other binaries to be.


> Where would sdcc binaries be?
`$(BINDIR)`

> Where would ihxcheck, bankpack, etc be?
`$(BINDIR)`

I just noticed, that should make the makefile place all binaries actually in `$(BINDIR)`, since they are still put tat `$(TARGETDIR)bin`. I gonna change that.

> Where would inc and lib be?
still at `$(TARGETDIR)lib/small/asxxxx/`, which can be changed with `--prefix=` and `--model-small`/`--model-medium`
and at `$(TARGETDIR)include` respectively

> How does it interact with set_gbdk_dir()?

I have to look into this. And see under which conditions this is set.

> Has it been tested on Windows? (Maybe not an issue if this is expected to be Linux only, but then should be noted)

No, I have no active Windows VM.

> And, could any of these be used as an alternative to this change?

Those are runtime flags. This PR is about compile time flags to set up the default directories. But if this compile time flag is set, it's impossible to change bin dir with `--prefix` later on, `BINDIR` did not include `%prefix%`.
It should be possible to get it working  by setting `--prefix=`, `--gbdklibdir` and `--gbdkincludedir`, but this would require you to run lcc through a wrapper script to work.

> Or, would adding an override for sdccdir be more flexible instead of compiling the alternative directory into the executable?

That is a completly different thing. Again it's a runtime flag (#133 actually tries to implement this). If implemented correctly you wouldn't be able to change the paths of bankpack and ihxcheck, because they should be independant of `%sdccdir`, since they are not from sdcc.

bbbbbr:
Thanks for the details, looks good to me.

At some point, probably a good idea to add a very short explanation of the purpose of BINDIR to both affected Makefile's at some point so that people in the future can understand it. :)

PR #148, opened 2021-02-28
Docs: remove getting started link until it's available after new docs

bbbbbr:


untoxa:
i think we should publish 4.0.3 instead

bbbbbr:
No preference :+1: I have a handful of things I'd like to update and add, but all of it can wait for another release. (such as large map sample, but it needs more work)

Issue #169, opened 2021-04-04
Updating some examples to have a better structure


Th3T3chn0G1t:
Heyo,

Opening this mainly as a tag target for PRs

Just some issues with examples which need to be addressed:
- The structure of the examples could be reworked - i.e. separate `src` and `obj` folders for building
- Some of the content is a obscure or unclear
- Some of the content contains outdated paradigms (https://github.com/gbdk-2020/gbdk-2020/issues/162)
- The structure of the `examples` folder could be reworked into more specific categories
- More broad examples implementing games or demos

I'll make a pr about some of the structural stuff in a bit!

bbbbbr:
@untoxa has suggested a couple things:
* For `src` only those example worth reorganizing, that contain several source files
* `obj` folder is not needed for one-line (lcc -o out.gb in1.c in2.c ...) compile examples since that does not leave any objects behind
* It is worth making new examples instead of tuning the old ones, we are lacking some basic examples
* Maybe reorganize the structure like: examples/games, examples/beginner, examples/hardware

Some of my thoughts:
* We can give people a better base to build from. Many people will just copy, use and build on what they find (and perhaps ignore the template projects), so having samples be more uniform a little better organized could save some trouble in long run. 
* Having `src` and `obj` or equivalents can help make the projects generally tidier
* Would be nice for the makefiles to generally harmonized and not have too many different variants
* Having comments in the makefile to explain what is being done and why is useful. I've been trying to do that in some of the more recent examples.
* GBDK Playground has some useful examples we may want to consider or be inspired by (may need updating to GBDK-2020 from gbdk 2.x): https://github.com/mrombout
* The scope for improving the example projects can be small or big :)

Th3T3chn0G1t:
Regarding the hardcoding of the compiler path in the example makefiles - might it be an idea to set an env variable at installation/tell the user to set an env variable pointing to the binaries. That way the examples can be used without modification whether gbdk binaries are installed to the `PATH` or otherwise

PR #170, opened 2021-04-04
Started restructuring examples - first 3 banking examples alphabetically

Th3T3chn0G1t:
Banking e.g. makefiles are still sub-par but better 

Addressing #169


gbdk-lib/examples/gb/banks_autobank/Makefile
```diff
@@ -1,30 +1,22 @@
-LCC	= ../../../bin/lcc -Wa-l -Wl-m
-BANKPACK  = ../../../bin/bankpack
+LCC	= lcc -Wa-l -Wl-m
```
bbbbbr, line 1: Same as comment above about keeping relative path

gbdk-lib/examples/gb/banks_autobank/Makefile
```diff
@@ -1,30 +1,22 @@
...
 
-BINS	    = autobanks.gb
-CSOURCES   := $(wildcard *.c)
-ASMSOURCES := $(wildcard *.s)
-
-# Compiling will produce .o files
-OBJS       = $(CSOURCES:%.c=%.o) $(ASMSOURCES:%.c=%.o)
-
-all:	$(BINS)
+all: obj $(BINS)
```
bbbbbr, line 7: It would be good to keep .asm sources in the list so that the makefile would still work if they wanted to base a project on it that included asm source files.

gbdk-lib/examples/gb/banks_autobank/Makefile
```diff
@@ -34,20 +26,11 @@ make.bat: Makefile
...
 # -Wl-ya4      : Use 4 RAM banks
 # -Wl-yt19     : Use MBC5 cartridge type
-$(BINS):	$(OBJS)
-	$(LCC) $(CFLAGS) -autobank -Wb-ext=.rel -Wb-v -Wl-yt19 -Wl-yoA -Wl-ya4 -o $(BINS) $(OBJS)
-
-
-# It's also possible to do all the compiling, autobanking and linking in a single call to lcc
-# with none of the other makefile "targets" above. You can try this target out by using "make onepass".
-# Notice that the input files are the .c and .asm sources instead of the compiled object files.
-onepass:
```
bbbbbr, line 44: The one-line alternate, example target should be kept. The purpose of the makefile is to help teach users what the possibilities are and how to use them.

gbdk-lib/examples/gb/banks_farptr/Makefile
```diff
@@ -1,34 +1,26 @@
-CC	= ../../../bin/lcc -Wa-l -Wl-m -Wl-j
+LCC	= lcc -Wa-l -Wl-m
```
bbbbbr, line 1: Why remove the noice output (-Wl-j)?

gbdk-lib/examples/gb/banks_autobank/Makefile
bbbbbr: Also probably good to add the no-ice output (-Wl-j) to harmonize it with other examples

gbdk-lib/examples/gb/banks/Makefile
```diff
@@ -1,63 +1,54 @@
-CC	= ../../../bin/lcc -Wa-l -Wl-m -Wl-j
+CC	= lcc -Wa-l -Wl-m -Wl-j
```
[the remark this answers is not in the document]
Th3T3chn0G1t: That's why I made a suggestion in the issue #170 about using an ENV variable for GBDK

gbdk-lib/examples/gb/banks_autobank/Makefile
Th3T3chn0G1t: I think given the inconsistency of the makefiles it might be an idea to just provide a boilerplate one rather than making them *all* fully featured

gbdk-lib/examples/gb/banks_autobank/Makefile
Th3T3chn0G1t: Same as above - perhaps just provide specific examples for makefiles

Th3T3chn0G1t: Though IDK since its all gonna continue being reworked at some point

gbdk-lib/examples/gb/banks_autobank/Makefile
Th3T3chn0G1t: Ah whoops - that might have just been me fat fingering it

PR #173, opened 2021-04-17
Work around segfault when using -Wl-yp

basxto:
`-Wl-yp`/`-Wm-yp` currently makes 4.1.0's makebin segfault, which results in lcc emitting:
> lcc: fatal error in /usr/bin//makebin

### DIFF
--- a/gbdk-support/lcc/lcc.c
+++ b/gbdk-support/lcc/lcc.c
@@ -860,8 +860,8 @@ static void opt(char *arg) {
 					sprintf(tmp, "%c%c%c", arg[3], arg[4], arg[5]); //-yo -ya -yt -yl -yk -yn -yp
 					if (!(arg[5] == 'c' || arg[5] == 'C' || arg[5] == 's'  || arg[5] == 'S' || arg[5] == 'j' || arg[5] == 'p')) // Don't add second arg for -yc -yC -ys -yS -yj
 						sprintf(tmp2, "%s", &arg[6]);
-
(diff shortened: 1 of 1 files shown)


bbbbbr:
For clarity, does this make LCC emit it with no space?
  `-yp0x<someval>=0x<someval>` 
instead of:
  `-yp 0x<someval>=0x<someval>`

You mentioned in discord that in the future makebin will support both styles, with and without spaces. Has this already been committed, and when is the estimate for the next release? If it is soon, what about just waiting for the makebin fix to be available?

basxto:
> For clarity, does this make LCC emit it with no space?
> `-yp0x<someval>=0x<someval>`
> instead of:
> `-yp 0x<someval>=0x<someval>`

Yes


> You mentioned in discord that in the future makebin will support both styles, with and without spaces. Has this already been committed, and when is the estimate for the next release? If it is soon, what about just waiting for the makebin fix to be available?

Nope, that's not fixed yet.
Next release is probably next year or so.

untoxa:
we don't care about releases much, we are trying to use stable snapshots. is that fixed in the current snapshots?

bbbbbr:
Looks like it's not fixed in the sdcc snapshots? At took a quick look and didn't see an explicit commit to fix it, but I could have missed it.

basxto:
It's not fixed in sdcc yet.

bbbbbr:
Since you indicate that future makebin will support this style, it seems ok.

PR #181, opened 2021-05-03
Png2mtspr : empty metasprites and fix for pngs over 255 pixels in height or width

bbbbbr:
* Turns off suppression of "blank" metapsrite frames (frames composed of entirely transparent sprites). 
* Fix endless loop bug for png files taller than 255 pixels (unsigned char used for x/y coordinates), also tidy a (int) to (long int) comparison warning and harmonize the types a little more (`int` is used most by function params, so that was used)

* Also add missing tools prefix for cross-compile support in Makefile (empty and not used by default)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ VER = 3.00
 
 # Prefix to add to the standard tools.  Usefull for a standard gcc
 # cross-compile.
+# TOOLSPREFIX = i686-w64-mingw32-
 TOOLSPREFIX =
 
 TARGETCC = $(TOOLSPREFIX)gcc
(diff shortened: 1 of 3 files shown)


untoxa:
>Turns off suppression of "blank" metapsrite frames (frames composed of entirely transparent sprites).

do they? i thought animation frame of zero sprites is possible.

bbbbbr:
> > Turns off suppression of "blank" metapsrite frames (frames composed of entirely transparent sprites).
> 
> do they? i thought animation frame of zero sprites is possible.

Display of them by the lib works, creation of them did not.

This code removed metasprites which did not contain any non-transparent sprites.
```
if(mt_sprite.size() == 0)
    sprites.pop_back();
```

Then, even if the empty metasprites had been allowed, the output would have failed to add the required `{metasprite_end}` terminator for empty ones.

PR #191, opened 2021-05-31
lcc: make extension handling code easier to understand

bbbbbr:
An attempt to make the extension handling code easier to understand and maintain, while still keeping the definition of extensions centralized. 

Also adds some comments where the code is still harder to read.

It's more readable, but still a little awkward. Feedback or ideas to improve it more?

For example:
From: `ofile = concat(base, first(suffixes[3]));`
To: `ofile = concat(base, EXT_O);`

From: `target_is_ihx = (suffix(outfile, suffixes, 5) == 4);`
To: `target_is_ihx = (suffix(outfile, (char *[]){EXT_IHX}, 1) != SUFX_NOMATCH);`

It doesn't entirely improve code like this: (`Count number of (.c, .i, .asm, .s) files`)
From: `else if (suffix(argv[i], suffixes, 3) >= 0)`
To: `else if (suffix(argv[i], suffixes, 3) != SUFX_NOMATCH)`

### DIFF
--- a/gbdk-support/lcc/gb.c
+++ b/gbdk-support/lcc/gb.c
@@ -243,6 +243,8 @@ static void buildArgs(char **args, const char *template)
 	*last = NULL;
 }
 
+// If order is changed here, file type handling MUST be updated
+// in lcc.c: "switch (suffix(name, suffixes, 5)) {"
 char *suffixes[] = {
     EXT_C,               // 0
     EXT_I,               // 1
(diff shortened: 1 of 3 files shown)


untoxa:
yes, suffix() is odd. can't even invent meaningful aliases for those values passed as a parameter for n.

bbbbbr:
I'll stop at minor improvements for now. The bigger fixes can wait until another day.

Issue #162, opened 2021-03-26
Why use custom data types (UINT16, etc)?


spth:
Why does gbdk use data types like UINT16, DWORD, etc?

As far as I can see, they are just equivalent to standard C types from stdint.h, so why not just use the latter?

Also, gbdk uses signed char as BOOLEAN. From the name I guess that type is meant to just represent boolean values, so bool from stdbool.h should be a more efficient choice.

untoxa:
Hello Philipp,

Friday, March 26, 2021, 11:59:24 AM, you wrote:

PKK> Why does gbdk use data types like UINT16, DWORD, etc?

PKK> As far as I can see, they are just equivalent to standard C types from stdint.h, so why not just use the latter?

PKK> Also, gbdk uses signed char as BOOLEAN. From the name I guess that type is meant to just represent boolean values, so bool from stdbool.h should be a more efficient choice.

that's legacy of the old GBDK design. i agree, that is a bad practice.


-- 
Best regards,
 Tony                            ***@***.***

bbbbbr:
Replaced `UBYTE`/`UINT8`/etc with stdint types for examples and gbdk libs where appropriate.

Does not include changes for `BOOLEAN`

untoxa:
i think we can close this now, due to: https://github.com/gbdk-2020/gbdk-2020/pull/195 and https://github.com/gbdk-2020/gbdk-2020/pull/196

PR #194, opened 2021-06-04
 Examples: convert to using stdint.h types instead of non-standard UINT8 / UBYTE / WORD etc Re: #162

bbbbbr:
Many users will follow the pattern set by the examples. It's been suggested (and I agree) that it's better to use standard types instead of custom ones.

(UBYTE|UINT8)  -> uint8_t 
(BYTE|INT8) -> int8_t 

(UWORD|UINT16) -> uint16_t 
(WORD|INT16) -> int16_t 

Samples all build and run.

Includes a couple updates for docs.

OK with this change?


untoxa:
we should convert library headers first because it is odd and confusing when examples do not match the api and docs.

PR #222, opened 2021-08-04
Add simpler way of accessing high and low bytes in "fixed" union type

bbbbbr:
- Allows using `.l` and `.h` directly for the `fixed` type without requiring `.b` as a prefix (`.b.h` and `.b.l` still work of course)

@Zal0 and @untoxa  - No worries if either of you aren't into this patch


untoxa:
This is fine

PR #240, opened 2021-08-28
Add ci/cd section with gbdk github action builder

wujood:
I created a github action for the gbdk-2020.
This could be a great addition to this documentation page.

### DIFF
--- a/docs/pages/02_links_and_tools.md
+++ b/docs/pages/02_links_and_tools.md
@@ -118,3 +118,11 @@ This is a brief list of useful tools and information. It is not meant to be comp
     Add line-by-line C source code to the main symbol file in a BGB compatible format. This allows for C source-like debugging in BGB in a limited way.
     https://gbdev.gg8.se/forums/viewtopic.php?id=710
 
+
+@anchor ci/cd
+# Continuous Integration/Deployment
+  - @anchor GBDK GitHub Action Builder
+    __GBDK GitHub Action Builder__  
+    Basic CI/CD capabilities for gbdk projects.  
+    https://github.com/wujood/gbdk-2020-github-builder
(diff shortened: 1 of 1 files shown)


bbbbbr:
Thanks, good idea. I've made a couple style and doxygen formatting edits.

bbbbbr:
By the way, I recently started looking into building GBDK itself with github actions. If you might be interested in helping out with that let us know. It seems straightforward so far, but having some more knowledge and experience could be helpful.

wujood:
Do you want to create a new GitHub action or a workflow? A workflow would be enough i guess

bbbbbr:
Yes, you're right just a workflow is suitable- the terminology and modern CI tools are new to me.
Here's the current work in progress.
https://github.com/bbbbbr/gbdk-2020/blob/build/.github/workflows/main.yml

At some point, in addition to periodic build packaging, we'd also like to run tests within the gbdk environment (create some specific tests, compile them using sdcc + gbdk, run in an emulator and capture output for validation)

wujood:
I would suggest the following:

- Create a workflow that builds the project on each push on every branch to validate the integrity of the changes.
- Create a workflow that builds and publishes the binaries if a release is created.
- In addition to the binaries you should create a docker image that has the latest binaries installed and is ready to compile gbdk projects.

That way developers can create workflows and GitHub actions while using the latest docker image. I could provide help if you help me create the scripts to actually build the binaries. 

I noticed that you had a workflow with a lot of if-statements for the different platforms. I would suggest to split them into different files if you have almost nothing in common to keep the overhead small and isolate bugs faster.

bbbbbr:
Let's continue this in a new issue: #241

PR #261, opened 2021-09-15
Examples: Crossplatform: Banks: Fix linux build

bbbbbr:
- switch to `sed` to generate the `-boN` `-baN` arguments from filenames (`-n ` and the trailing `/p` suppress output when no match is found)
- make naming scheme more literal `somefile.ba1.bo3.c` so that it's easier to follow


bbbbbr:
Doesn't seem to work on macOS, need to look into whether it has `sed` by default and whether it's compatible with `-n` and `/p`

PR #294, opened 2021-12-28
Fix letter U appearing as K

splch:
This is how the `font_ibm` loads:
image
But here is the `font_min`:
image

21 correlates with the letter K, which renders correctly; however, 21 is also assigned to the position where U should appear.

This pull request replaces the repeated letter K (21) with U (31).


untoxa:
nice catch, thank you.

PR #298, opened 2022-01-09
Include CGB in Supported Consoles

splch:
Add Game Boy Color to line with Game Boy support. Also reorder the listing to reflect that of the [docs](docs/pages/06b_supported_consoles.md) page more accurately.


bbbbbr:
Good point to add GBC to the list

Since this list isn't in tree format like the docs are it's better to keep Megaduck at the bottom since it's the least important / likely to be used

splch:
> Good point to add GBC to the list
> 
> Since this list isn't in tree format like the docs are it's better to keep Megaduck at the bottom since it's the least important / likely to be used

That makes sense. I reverted the ordering change in the file :smile:

PR #299, opened 2022-01-09
Make buffer[] a local variable in init()

splch:
This should automatically deallocate buffer from memory once it's out of scope. I also mentioned that the size was determined from the gbcompress tool since GBTD doesn't give that information.


untoxa:
that is just an example. allocating such big structures on stack may cause even more troubles than allocating them statically.

PR #323, opened 2022-02-15
Create compile.bat

DaSalba:
Add a Windows batch file to compile all examples in one go, just as you can with the Linux Makefile.

### DIFF
--- a/gbdk-lib/examples/ap/compile.bat
+++ b/gbdk-lib/examples/ap/compile.bat
@@ -1,0 +1,9 @@
+@echo off
+
+REM Automatically build all Analogue Pocket examples in this directory.
+
+for /d %%i in (.\*) do (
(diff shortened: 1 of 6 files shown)


bbbbbr:
Can you add this for the other example folders as well?

DaSalba:
Sure, it's done. Though it would be ideal if we could generate them automatically from the Makefiles, so they would only appear in the releases, not the repo, but I haven't managed to do it.

bbbbbr:
Auto-generating them in the build output could be done, but I don't think there is much improvement from it. It's ok for them to exist alongside the Makefiles.

One of the main reasons the per-example compile.bat files are auto-generated is because the examples are subject to change. These compile.bat files look like they should remain universal regardless of example folder changes.

PR #329, opened 2022-03-06
adding changes for source tileset, binary output, and transposing map…

LaroldsJubilantJunkyard:
Added Arguments:

- **-bin** for binary output.
- **-transposed** for column-by-column output instead of row-by-row
- **-source-tileset** Which takes in another PNG image to use as the "tileset". the current image being passed into png2asset would be the "tilemap".
- **-tiles_only** prevents output of maps/metasprites
- **-metasprites_only** only outputs metasprites
- **-maps_only** only outputs the maps and attributes
 
Also added a palette count constant for convenience.

### DIFF
--- a/gbdk-support/png2asset/png2asset.cpp
+++ b/gbdk-support/png2asset/png2asset.cpp
@@ -4,6 +4,7 @@
 #include <cstring>
 #include <set>
 #include <stdio.h>
+#include <fstream>
 #include "lodepng.h"
 
 using namespace std;
@@ -126,6 +127,14 @@ struct MTTile
 	MTTile(char offset_x, char offset_y, unsigned char offset_idx, unsigned char props) : offset_x(offset_x), offset_y(offset_y), offset_idx(offset_idx), props(props) {}
 	MTTile() : offset_x(0), offset_y(0), offset_idx(0), props(0) {}
 };
+string source_tileset;
+int extra_tile_count = 0;
+int source_palette_count = 0;
+int source_tileset_size = 0;
+bool includeTileData = true;
+bool includedMapOrMetaspriteData = true;
+PNGImage source_tileset_image;
+bool use_source_tileset = false;
 
 typedef vector< MTTile > MetaSprite;
 vector< Tile > tiles;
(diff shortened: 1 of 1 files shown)


bbbbbr:
* One of the most important things is to make sure it still works with the ZGB. Maybe download it, build the template and then swap your version in to see if there are problems. Would like to see that confirmed before merge.

**Overall the code looks clean, decently commented and easy to understand. The features look good.**

* I cloned, built on linux and tested with a bunch of the existing examples (metasprites, logo). In the brief test everthing looked good. I won't have time for much testing of the new features myself (hence below about a new example project), so I'll assume that they work as expected. :)

* (Can be separate PR, or same) Would be nice if to also have a concise project in the examples folder that exercised some (or all) of the new settings. That would demonstrate how to use them, and to verify they work as expected.

* (Can be separate PR, or same) You're welcome submit changes to the docs to cover these as well, or if you give me skeleton descriptions I can expand them. https://gbdk-2020.github.io/gbdk-2020/docs/api/docs_toolchain.html#autotoc_md109

Some minor stuff:

* These look like debug printings that might be trimmed down if they are going to print for every extra tile. Might be better if they were just a summary at the end.

printf("found a tile not in the source tileset at %d,%d. The target tileset has %d extra tiles.\n", x, y,extra_tile_count+1);

https://github.com/gbdk-2020/gbdk-2020/pull/329/files#diff-2f04301949ee791977fd96f370134aaead432ebb9c610b7b4a3d1a732dea7784R242
https://github.com/gbdk-2020/gbdk-2020/pull/329/files#diff-2f04301949ee791977fd96f370134aaead432ebb9c610b7b4a3d1a732dea7784R279


* Suggestion:
printf("-source_tileset     source tileset\n");
printf("-source_tileset     use a separate source tileset (in png format)\n");

* Suggestion:
printf("-bin				Should a .bin file be written?");
printf("-bin				write output as binary instead of C format");

* Suggestion:
printf("-transposed         Should be map items be written column-by-column, instead of row-by-row?");
printf("-transposed         write map items out column-by-column instead of row-by-row);

* Typo "The the":
printf("error: The the number of color palette's for your source tileset (%d) and target image (%d) do not match.", source_tileset_image.palettesize, image.palettesize);

https://github.com/gbdk-2020/gbdk-2020/pull/329/files#diff-2f04301949ee791977fd96f370134aaead432ebb9c610b7b4a3d1a732dea7784R769


* Palettes: 
   - Just thinking: handling is probably good for now. In the long run we may want the ability for additional palette entries beyond what is in the source tileset. For example, a source tileset that only has 4 CGB palettes and leaves the rest for the current map to alter.
   -  The requirement should be documented
printf("error: The palette's for your source tileset and target image do not match.");

https://github.com/gbdk-2020/gbdk-2020/pull/329/files#diff-2f04301949ee791977fd96f370134aaead432ebb9c610b7b4a3d1a732dea7784R769

* Long term: As the code grows, one of these days the pngasset source should start to get split up into separate files so it is easier to navigate.

PR #344, opened 2022-05-26
Link to relevant pandocs MBC section

splch:
I think it's useful to link to the section in markdown. Would that be okay despite it not being linked in the actual man page?


bbbbbr:
It is ok to add the link, although it runs the risk of becoming obsolete if they change the way they render the site.

However, it cannot be changed in the file you have PR'd. The `docs/20_toolchain_settings.md` file is auto-generated from the help output of each individual tool. (via `make docs-toolchain-generate`. Please do not PR the output of that though, we will do that when closer to next release)

Instead it has to be changed in bankpack itself: 
https://github.com/gbdk-2020/gbdk-2020/blob/develop/gbdk-support/bankpack/bankpack.c#L36

If you make that change, please confirm that bankpack successfully builds after the modifications.

splch:
I see :) , I confirmed that the docs built and have updated the PR to only modify `bankpack.c`

bbbbbr:
Thanks, I've split the new content to a separate line, merging now

PR #352, opened 2022-06-04
Issue#351: Fix linker error when using -u parameter to create .rst files from .lst files

michel-iwaniec:
* Update gbdk-lib-install-platforms target to copy crt0.lst along with crt0.o

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,7 @@ gbdk-lib-install-platforms:
 		echo Installing lib for platform: $$plat; \
 		mkdir -p $(BUILDDIR)/lib/small/asxxxx/$$plat; \
 		cp $(GBDKLIBDIR)/build/small/asxxxx/$$plat/crt0.o $(BUILDDIR)/lib/small/asxxxx/$$plat/crt0.o; \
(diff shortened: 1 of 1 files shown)


untoxa:
i think this is actually better: https://github.com/gbdk-2020/gbdk-2020/commit/24fc5273c0cd4838a8d538f0602489cd1995ae30

michel-iwaniec:
> i think this is actually better: [24fc527](https://github.com/gbdk-2020/gbdk-2020/commit/24fc5273c0cd4838a8d538f0602489cd1995ae30)

I quite disagree on that statement - it's an inferior solution, as it results in an empty / useless .rst file of zero size for crt0.

But it's your repo at the end of the day.

untoxa:
I think empty list file is better, because we don't ship the library sources with the dist. If we put the actual crt0.lst, then why not for the whole library then? So, currently the idea is, that the user may get rst files for the sources he wrote. The other problem with rst-files, is that the crt0.rst is created beside the crt0.o object file in the gbdk-2020 internal folders.

So, two additional problems here: 
1) you might not have rights for that 
2) for the each project the actual rst file (if being made from the actual non-zero crt0.lst file) may differ for each compilation for each project, which is not good.

PR #356, opened 2022-06-04
Issue#355: Fix bug with letter 'U' in f_min font appearing as 'K' for z80 targets, and eliminate file duplication

michel-iwaniec:
* Move font data files from gbdk-lib/libc/targets/sm83/ to gbdk-lib/libc/asm/
* Delete font data files in gbdk-lib/libc/targets/z80/
* Add font data files to gbdk-lib/libc/asm/<PORT>/Makefile
* Remove data files from gbdk-lib/libc/targets/<PORT>/<PLATFORM>/Makefile
* Remove unused loading code from f_ibm_full.s, so that the file only contains platform-independent data


untoxa:
why did you move those font files into libc/asm/ but not into libc/targets/ ? libc/asm/ is for the stuff that is processor-dependent, while libc/targets/ is for platform dependent. font subsystem is a platform thing, not processor. also there might be platform-dependent fonts, because format support platform-native tiles, not only 1bpp font. please move those fonts into libc/targets/

michel-iwaniec:
> why did you move those font files into libc/asm/ but not into libc/targets/ ? libc/asm/ is for the stuff that is processor-dependent, while libc/targets/ is for platform dependent. font subsystem is a platform thing, not processor. also there might be platform-dependent fonts, because format support platform-native tiles, not only 1bpp font. please move those fonts into libc/targets/

I wasn't sure which directory was more suitable, so just went with this one. I'll move them and push a new commit to this PR shortly.

PR #360, opened 2022-06-07
Fix build system bug accidentally introduced by PR#356

michel-iwaniec:
* Strip all relative paths from font filenames in each targets/CPU/Makefile
* Add additional rule for stripping ../../ from .o filename in rules-asxxxx.mk


untoxa:
Thank you, missed that.

PR #387, opened 2022-07-01
Docs: AP: more detail about the differences

bbbbbr:


bbbbbr:
Says merge failed, but it looks like it actually went through in: 
46d2b5d449cd3b66bdd2e165f82b87b134b5afd9

PR #391, opened 2022-07-10
Optimize controller-reading routine

ISSOtm:
Adapted from https://github.com/pinobatch/libbet/blob/a01e21eee6ac32fa9dc250effd74e8ac5eeb3b97/src/pads.z80#L61-L95, with the license blurb correctly preserved.

### DIFF
--- a/gbdk-lib/libc/targets/sm83/pad.s
+++ b/gbdk-lib/libc/targets/sm83/pad.s
@@ -27,31 +27,56 @@ _waitpadup::
 	;;   0x40 - Select  0x04 - Up
 	;;   0x20 - B	    0x02 - Left
 	;;   0x10 - A	    0x01 - Right
+;
+; Controller reading for Game Boy and Super Game Boy
+;
+; Copyright 2018, 2020 Damian Yerrick (adapted by Eldred Habert)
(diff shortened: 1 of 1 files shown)


bbbbbr:
Can you briefly describe the improvements this offers?

And, could it be re-licensed to be under the GPLv2 + LE so that its license is in harmony with the majority the library? Given what we have just gone through with licensing it would be ideal to not accumulate more patchwork in that area.

bbbbbr:
A couple other things:
When building with this patch there is an error (syntax format):
> `../pad.s:78: Error: <a> Invalid Addressing Mode.`

If that is fixed then there is an additional error when linking to it in crt0.s:
> `?ASlink-Warning-Undefined Global '.1$' referenced by module ''
> ?ASlink-Warning-Undefined Global '.2$' referenced by module ''

Once those are resolved I did a rough measurement and it seems to be slower than the original code:
This PR: ~356 cycles
Original code: ~220 cycles

untoxa:
it is slower to match the recommended timings; also it waits the equal time for both: d-pad and buttons, it is not clear from the doc, whether delays should be the same or different.

PR #449, opened 2022-11-14
Examples: Sound test: MegaDuck support, init wave ram

bbbbbr:


untoxa:
hmmm... `nybble_swap_if_megaduck()`? suggest more generic names for that, like `platform_translate()`.

bbbbbr:
Can change the naming if you'd prefer.

The reasoning for the names was:
- That program is very specific to the Game Boy sound hardware. Unlikely it is ported to a non-GB like console.
- The MegaDuck is the only known GB clone with register variations
- Choose function names which emphasizes what those differences are for easy understanding.
  - `nybble_swap_if_megaduck()`
  - `NR32_swap_volume_if_megaduck()`

untoxa:
translate_volume(), translate_envelope(). why you put megaduck there? what if we need more than 1 ifdef there in future?

bbbbbr:
I can rename them to be more general purpose, but the argument for it is not persuasive.

> why you put megaduck there? what if we need more than 1 ifdef there in future?

There is only one single known GB clone with register variations that exists: MegaDuck. What other purpose is anyone going to use an ifdef in there for?

untoxa:
there could be two with analogue pocket though :)

bbbbbr:
Let me rephrase that more specific - Analogue Pocket does not have Audio Hardware register address and bit flag changes.What would the ifdef for the sound hardware do?

PR #454, opened 2022-11-18
GB: Fix switched-around naming of move_metasprite_hflip / move_metasprite_vflip

michel-iwaniec:
* Exchange names for ___move_metasprite_hflip and ___move_metasprite_vflip in metasprites_flip.s
* Exchange names for move_metasprite_hflip and move_metasprite_vflip in gbdk-lib/include/gb/metasprites.h, make them call the right asm routines, and edit docs


michel-iwaniec:
Unfortunately would breaks backwards compatibility, so old naming may have to stick for now.

PR #462, opened 2022-12-03
Add sub-palette switching to cross-platform/metasprites example (for GBC and NES)

michel-iwaniec:
* Extend rot variable to switch between palettes after every X/Y/XY flip rotation, and apply to move_metasprite_* calls
* Setup pink, cyan and green palettes for GBC/NES to visualize switching
* Enable gbc as target in Makefile


untoxa:
i think palettes should come from the PNG itself?

michel-iwaniec:
> i think palettes should come from the PNG itself?

Usually, but not in this case. It's meant to test the (recently introduced) base_props. And changing the sub-palette to denote different variants of enemies is also a common cheap trick in games. (look at Zelda1 and its red/blue Octoroks)

PR #463, opened 2022-12-08
Cross-platform metasprite flipping support:

michel-iwaniec:
* Change sms_metasprites to have 4 permutations for flipping combinations, using a generic macro for function bodies
* Change examples/cross-platform/metasprites to duplicate uploaded tiles by X and Y depending on HARDWARE_SPRITE_CAN_FLIP_X/Y defines
* Change examples/cross-platform/metasprites to use 8x8 sprites, to make all duplicated tiles fit tile memory on sms/gg
* Remove hide / show metasprite functionality in examples/cross-platform/metasprites
* Move rotation to A button, to make compatible with 2-button SMS joypad


untoxa:
should not it be not just `neg`, but `neg/inc a` ?

PR #464, opened 2022-12-09
Eliminate code duplication in sms metasprites code:

michel-iwaniec:
* Put MOVE_METASPRITE_BODY into an include file sms_metasprites_macro.s
* Have each flip combination (noflip/flipx/flipy/flipxy) include and use the macro


bbbbbr:
Is this PR still applicable/relevant?

PR #465, opened 2022-12-09
Code Review changes to cross-platform/metasprites

michel-iwaniec:
* Make reverse_bits lookup table a const array
* Move tile upload/flipping into its own function load_and_duplicate_sprite_tile_data
* Replace joypad_ex polling with simpler joypad routine, as no need for 2-player support


bbbbbr:
Looks good, thanks!

PR #484, opened 2023-03-02
Update 01_getting_started.md for macOS users

ukd1:
Add instructions how to un-quarantine unsigned binaries for macOS users. Also fixed `MacOS` -> `macOS` to match Apple styling / branding.


bbbbbr:
Thanks for adding this, glad to see it's a very straightforward fix.

I'm going to generalize the instructions a bit to work in cases where `$GBDKDIR` has not yet been defined for the system, since some users may not have set that up yet (or may never set it up). Will probably also add a FAQ entry linking to it.

PR #486, opened 2023-03-12
Updated documentation for set_sprite_prop()

Adam-Shea:
I have updated the docs for set_sprite_prop() as they were a little bit out.
I'm quite interested in updating some other documentation, would following updates be accepted? Is there any preferred format?

### DIFF
--- a/gbdk-lib/include/gb/gb.h
+++ b/gbdk-lib/include/gb/gb.h
@@ -1749,6 +1749,19 @@ inline uint8_t get_sprite_tile(uint8_t nb) {
     \li Bit 1 - See bit 0.
     \li Bit 0 - GBC only. Bits 0-2 indicate which of the 7 OBJ colour palettes the
               sprite is assigned.
+
+    It's recommended to use GBDK constants (eg: S_FLIPY) to configure sprite properties as these are crossplatform.
+
+    \code{.c}
+    // Load palette data into the first palette
+    set_sprite_palette(4, 1, exampleSprite_palettes)
+    // Set the OAM value for the sprite
+    // These flags tell the sprite to flip both vertically and horizontally.
+    set_sprite_prop(0, 0b01500000);
+    set_sprite_prop(0, S_FLIPY + S_FLIPX);
+    \endcode
+
+    @see S_PALETTE, S_FLIPX, S_FLIPY, S_PRIORITY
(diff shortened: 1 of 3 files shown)


bbbbbr:
Looks ok to me, and yes more help on the documentation would be welcome. 

For format: if you mean what method of contributing to the docs, doing so as doxygen compatible PRs is suitable. With edits grouped together so there aren't too many PRs, but it's ok to divide them into categories. 

CC'ing @michel-iwaniec in for additional review as the main person on NES support.

We should probably update the docs for set_sprite_prop() to mention the constants below (when applicable), as they're cross platform and better than using the flag bits directly.
`@see S_PALETTE, S_FLIPX, S_FLIPY, S_PRIORITY`

untoxa:
I have a suggestion for the bit fields. Don't address bits absolutely, but always use definitions instead. Like S_PRIORITY and so on.

untoxa:
Oh, i see @bbbbbr already suggested that. Not only sprites, but also STAT and LCDC regs.

untoxa:
i mean remove absolute bit numbers at all. because some other platforms like megaduck or analogue pocket may have those bits scrambled.

Adam-Shea:
Should it not at least be mentioned that you could use absolute bit numbers although discouraged? When reading through the docs I found the absolute numbers easier to understand initially. Perhaps I should restructure the text to give using constants prority and then mention that absolute bits can also be used?

untoxa:
absolute numbers make code unreadable as well

bbbbbr:
Yes, it would be nice to have the details for each target platform (and noting they are platform specific).

But I don't think the absence for SMS/GG or MSX should block improving parts of NES docs on their own. Documenting can take time and often gets done in gradual pieces.

Adam-Shea:
I'm happy to update the docs for set_sprite_prop() for all of the systems in this pull request if that would help clear up this issue?

untoxa:
priority must be given to the cross-platform docs. and only then - platform specific. definitions are cross-platform. absolute bit numbers are not.

bbbbbr:
If the absolute bit examples are removed then we can merge this 
(`set_sprite_prop(0, 0b01500000);` and `set_sprite_prop(0, 0b11000000);`)

PR #488, opened 2023-03-26
Fix gbdk-lib-install-ports target dependencies

charitybell:
The gbdk-lib-install-ports has an unwritten dependency on the gbdk-lib-build target. the gbdk-lib-install-ports target fails if built before the gbdk-lib-build target. In a non-parallelized build, these two targets are built in the right order and the build works fine. In a parallelized build, a race condition causes the targets to build in the wrong order.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,7 @@ gbdk-lib-install: gbdk-lib-install-platforms
 gbdk-lib-install-prepare:
 	@rm -rf $(BUILDDIR)/lib
 
-gbdk-lib-install-ports:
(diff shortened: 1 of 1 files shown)


bbbbbr:
I haven't checked, so maybe you have - is it going to be a similar case for the `gbdk-lib-install-platforms` and `gbdk-lib-install-examples` targets?

charitybell:
I tried building with 2 jobs all the way up through 9 jobs, and all of them succeed, so I think the other targets are fine.

it would probably be a good idea to look through the makefile and fix any other targets that might have unwritten dependencies.

PR #489, opened 2023-04-11
Fix GG Header for some Emulators

RetroGamer02:
Some emulators read the Header as SMS-Export and set it into SMS Mode.


bbbbbr:
Thanks for proposing this, however it looks simpler to pass `-xj 6` to `makebin` when building for Game Gear.

That changes the region code with no code changes required for `makebin`.

PR #513, opened 2023-06-02
NES: Merge bank switching changes to sdcc_4.3

michel-iwaniec:


bbbbbr:
Congrats :+1: 
Gave it a look through and seems good to me. Per other mediums, builds ok and resulting binaries seem to work.

Minor suggestion for now or some time before release:

NES platform flag could be as below: It would be shorter and more easily readable as NES (I know it has the "ATM Machine" acronym word repeat thing)
  - `NINTENDO_ENTERTAINMENT_SYSTEM` -> `NINTENDO_NES`

untoxa:
yes, i agree, `NINTENDO_ENTERTAINMENT_SYSTEM` is a bit too verbose,

michel-iwaniec:
> Congrats 👍 Gave it a look through and seems good to me. Per other mediums, builds ok and resulting binaries seem to work.
> 
> Minor suggestion for now or some time before release:
> 
> NES platform flag could be as below: It would be shorter and more easily readable as NES (I know it has the "ATM Machine" acronym word repeat thing)
> 
> * `NINTENDO_ENTERTAINMENT_SYSTEM` -> `NINTENDO_NES`

Yeah, I agree it would be better.

It might also be a time to revisit other target defines... because "#ifdef NINTENDO" meaning "we are running on a GameBoy" is a bit funny now that we'll have multiple NINTENDO platforms :)

Not sure what's the best way to organise it though, considering we have some targets that are very similar but different hardware (gb/gbc/ap and gg/sms).

bbbbbr:
Agreed,  `NINTENDO` meaning Game Boy will be confusing with NES support becoming more prominent. 

While we should be careful with making breaking changes, I would be ok with changing it now before it gets more widely used. 


# Some ideas
 
- `GAMEBOY_CPU` or `GAMEBOY_LIKE`
- More obscure, but more specific: or`SM83_CPU` or  `GAMEBOY_PLATFORM`

# Current configuration
- `NINTENDO`
- GB: `GAMEBOY`  (from `__TARGET_gb`)
- GBC: `GAMEBOY` (from `__TARGET_gb`)
- AP: `ANALOGUEPOCKET` (from`__TARGET_ap`)
- DUCK: `MEGADUCK` (from `__TARGET_duck`)
---
`SEGA`
- SMS: `MASTERSYSTEM` (from `__TARGET_sms`)
- GG: `GAMEGEAR` (from __TARGET_gg)

---
- NES: `NINTENDO_NES` (from `__TARGET_nes`)
---
 
- MSXDOS: ``MSXDOS`` (from `__TARGET_msxdos`)

PR #519, opened 2023-06-07
Platformer Example PR into main repo

LaroldsJubilantJunkyard:
This example includes 3 different levels from 2 worlds.
png2asset is used in the makefile with 2 different world tilesets.
tiled ".tmx" files are included. They need to be exported as PNGs when changed (explained in readme)
When the player gets to the end, the example goes to the next level. 
the example cycles between the 3 levels.
The player can walk using the d-pad, run by holding B, and jump using A.
Code from the large map demo used for camera scrolling.
RAM variables are used instead of directly referencing level data.
Scaled integers are used for smooth motion.


untoxa:
Can this be compiled for the game gear as well?

LaroldsJubilantJunkyard:
> Can this be compiled for the game gear as well?

I'm not familiar with such. So, That'll require some additional effort and research.

PR #526, opened 2023-06-15
Avoid missing-dll-file error during build

michel-iwaniec:
* Remove copy of missing Windows .dll libgcc_s_dw2-1.dll from top-level Makefile


bbbbbr:
That DLL is needed for Windows 32 (and only win32). It's ok for now that it fails to copy since it doesn't cause Make to fail, but it is a little sloppy. If we can detect between win32 / win64 then we can have the Makefile choose the right dll conditionally.

https://github.com/gbdk-2020/gbdk-2020/blob/develop/Makefile#LL280C1-L282C38
```
# Copy SDDC executable files
# win 32 specific: libgcc_s_dw2-1.dll
# win 64 specific: libgcc_s_seh-1.dll
```

PR #529, opened 2023-06-25
Example program idea: Program that detects which Game Boy the ROM is being run on

veganlies2me:
The makefile should work. If not, tell me.

### DIFF
--- a/gbdk-lib/examples/gb/gbcheck/Makefile
+++ b/gbdk-lib/examples/gb/gbcheck/Makefile
@@ -1,0 +1,12 @@
+CC	= ../../../bin/lcc
+
+all:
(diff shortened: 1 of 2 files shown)


bbbbbr:
Two more changes:

To remove the warning ` warning 283: function declarator with no prototype`
- `void main()` -> `void main(void)`

To fix the Makefile
- `$(CC) -o -Wm-ys gbcheck.gb gbcheck.c ` -> `$(CC) -Wm-ys -o gbcheck.gb gbcheck.c `

veganlies2me:
Does it work now?

bbbbbr:
Thanks, the Makefile works now.

Did some testing on hardware and looked good.

PR #559, opened 2023-07-29
Update NES section explaining setting of attributes in depth

michel-iwaniec:


bbbbbr:
Looks good to me :+1:, thanks!

I might drop some of those details in the related functions if they aren't already there too - can't remember off the top of mind.

PR #564, opened 2023-08-06
Bugfix: skip keep_duplicate_tiles logic when using a source_tileset

dekstop:
This fixes a bug in the interaction between two features: when both source_tileset and keep_duplicate_tiles are used, the resulting generated map is just a list of steadily increasing numbers.

To reproduce:

* Create tileset.png and map1.png
* Call png2asset as follows: `png2asset -keep_duplicate_tiles -map -source_tileset tileset.png map1.png -c map1.c`

The resulting map will look as follows, regardless of the tile & map content:

```
const unsigned char map1_map[...] = {
	0x00,0x01,0x02,0x03,0x04,0x05,0x06,... (etc)
};
```

... because the keep_duplicate_tiles logic is applied to both the source tileset PNG *and* the map PNG, when it should only be applied to the former.

### DIFF
--- a/gbdk-support/png2asset/png2asset.cpp
+++ b/gbdk-support/png2asset/png2asset.cpp
@@ -184,7 +184,7 @@ void GetMetaSprite(int _x, int _y, int _w, int _h, int pivot_x, int pivot_y)
 				unsigned char props;
 				unsigned char pal_idx = image.data[y * image.w + x] >> 2; //We can pick the palette from the first pixel of this tile
 
-				if(keep_duplicate_tiles)
+				if((!use_source_tileset) && keep_duplicate_tiles)
 				{
 					tiles.push_back(tile);
 					idx = tiles.size() - 1;
@@ -233,7 +233,7 @@ void GetMap()
 			size_t idx;
 			unsigned char props;
 
-			if(keep_duplicate_tiles)
+			if((!use_source_tileset) && keep_duplicate_tiles)
 			{
 				tiles.push_back(tile);
 				idx = tiles.size() - 1;


dekstop:
Attaching a simple test case, tileset.png and map1.png, for use with the command line example given above.

tileset.png: 


map1.png: 


---

Expected tile map code in map1.c:
```
const unsigned char map1_map[16] = {
	0x00,0x00,0x00,0x00,
	0x02,0x02,0x02,0x02,
	0x01,0x01,0x01,0x01,
	0x03,0x03,0x03,0x03,
};
```

Actual tile map:
```
const unsigned char map1_map[16] = {
	0x04,0x05,0x06,0x07,
	0x08,0x09,0x0a,0x0b,

	0x0c,0x0d,0x0e,0x0f,
	0x10,0x11,0x12,0x13,
};
```

bbbbbr:
Thanks for opening the ticket.

Looks like this patch resolves the issue by locking out `keep_duplicate_tiles` if `keep_duplicate_tiles` is enabled. If we go this route, it would be helpful to have a comment briefly explaining the reason for the lockout.

Off the top of my head I'm unsure whether we might want to allow using those two features together, so I'll check into that. Doesn't look too hard to do that if needed (only performing `FindTiIe()` on tiles from the source tileset, which aren't currently partitioned from tiles of the image being converted... so limit the search to the first N (`source_tileset_size`) tiles).

dekstop:
Thanks for checking in! 

Just to clarify (I should have done this in my PR) -- the patch does not actually disable the keep_duplicate_tiles feature in such cases. It only ensures that when both keep_duplicate_tiles and source_tileset are used, keep_duplicate_tiles is only applied to the source tileset, not to the map. I think that this is the desired behaviour in all cases when both features are used. 

This outcome is a result of how source_tileset is implemented: it calls GetMap in two passes, first for the tileset, then for the map; but disables "source_tileset" for the first call. See here in GetSourceTileset: https://github.com/gbdk-2020/gbdk-2020/blob/0d0b40415db18c231ff8ad7177c2aed522cd3022/gbdk-support/png2asset/png2asset.cpp#L624

My current workflow relies on the combination of both features, so I would prefer they can coexist. (Specifically, it is useful for works in progress where a shared tileset still has some empty tiles when developing the first maps, but when references to specific tile numbers will be used for collision detection.)

bbbbbr:
OK, perfect! Thanks for the additional details. I'll merge

PR #592, opened 2023-10-30
Feature/layered metasprites

LaroldsJubilantJunkyard:
After a tile's palette is recorded, whenever another the palette for a pixel is different, it will save that data on a copy of the metasprite. Eventually, we export tiles per palette.

### DIFF
--- a/gbdk-support/png2asset/metasprites.cpp
+++ b/gbdk-support/png2asset/metasprites.cpp
@@ -18,7 +18,7 @@
 
 using namespace std;
 
-void GetMetaspriteLayeredTiles(Tile& tile, MetaSprite& mt_sprite, int* last_x, int* last_y, int x, int y, PNG2AssetData* assetData) {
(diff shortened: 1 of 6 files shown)


bbbbbr:
Migrating some old discord comments to the PR itself so they're easier to keep track of. Not sure how many of them have been addressed.

-----
Thanks for PRing this, could be a cool feature.
- Seems like it should be gated by an option and off by default, otherwise people may (often?) accidentally use more sprites than intended. Maybe @untoxa and @michel-iwaniec have opinions about this?
- In metasprites.cpp:  `GetMetaSprite()`: The max sprites per line test and warnings at the end increase the function size a lot, they could be in a sub function to reduce that and make it more readable
- Might be nice to see brief descriptions above the new functions indicating what their role in the processing pipeline 
- Would be good to have a cross platform sample project that both demos the feature and can serve as a basic test for it

PR #606, opened 2024-01-03
Enable verbose debug flags in examples:

michel-iwaniec:
* -v (lcc verbose output)
* -Wa-l (.lst file from assembler)
* -Wl-u (.lst -> .rst address update)
* -Wl-w (wide listing in .map file)


bbbbbr:
See #614 instead

PR #611, opened 2024-01-15
moving plain pointers to std::vector

SelvinPL:
also fixing memory leak

### DIFF
--- a/gbdk-support/png2asset/image_data.cpp
+++ b/gbdk-support/png2asset/image_data.cpp
@@ -68,15 +68,15 @@ int ReadImageData_KeepPaletteOrder(  PNG2AssetData* assetData, string input_file
     // So: state->info_png.color.palette/size instead of state->info_raw.palette/size
     unsigned int palette_count = PaletteCountApplyMaxLimit((unsigned int)assetData->args->max_palettes, (unsigned int)(state.info_png.color.palettesize / assetData->image.colors_per_pal));
     assetData->image.total_color_count = palette_count * assetData->image.colors_per_pal;
+    auto new_palettes_size = assetData->image.total_color_count * RGBA32_SZ;
     // Copy the palette data since the source buffer (state...) won't exist outside the scope of this function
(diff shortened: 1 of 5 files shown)


gbdk-support/png2asset/image_data.cpp
```diff
@@ -68,15 +68,15 @@ int ReadImageData_KeepPaletteOrder(  PNG2AssetData* assetData, string input_file
     // So: state->info_png.color.palette/size instead of state->info_raw.palette/size
     unsigned int palette_count = PaletteCountApplyMaxLimit((unsigned int)assetData->args->max_palettes, (unsigned int)(state.info_png.color.palettesize / assetData->image.colors_per_pal));
     assetData->image.total_color_count = palette_count * assetData->image.colors_per_pal;
+    auto new_palettes_size = assetData->image.total_color_count * RGBA32_SZ;
```
SelvinPL, line 71: can be replaced with `size_t`

gbdk-support/png2asset/image_utils.cpp
```diff
@@ -100,10 +100,10 @@ static int pixel_find_color_in_palette(const PNGImage& image, int x, int y, int
 
     // Get pixel palette index num, then get RGB value for it
     uint8_t pixel_pal_idx = image.data[(y * image.w) + x];
-    uint8_t * p_pixel_color = &image.palette[pixel_pal_idx * RGBA32_SZ];
+    auto p_pixel_color = &image.palette[pixel_pal_idx * RGBA32_SZ];
```
SelvinPL, line 103: can be replaced with `const uint8_t*`

gbdk-support/png2asset/image_data.cpp
bbbbbr: Sure

gbdk-support/png2asset/image_utils.cpp
bbbbbr: Sure

PR #614, opened 2024-01-17
Debug improvements: more flags for lcc -debug, GBDK_DEBUG env var enable for examples

bbbbbr:
- lcc: add to -debug the following flags: -Wa-l -Wl-u -Wl-w
- Examples: Add GBDK_DEBUG environment var to easily turn on and off debug flags
- Improve up clean target file type coverage for a couple examples


bbbbbr:
This should also address #606 while keeping the default output more compact unless requested.

PR #621, opened 2024-02-01
Refactor png2assets arguments for logo example

brettinman:
Hi, I encountered some issues figuring out how to build a logo ROM, and wanted to contribute back to this project by refactoring this to make it easier for future beginners.

Ultimately what needed to happen was adding -noflip to the pn2asset arguments list. However, those were stored in .meta files that could not include comments, so there didn't seem to be a good solution on how I could add this hint.

So what I chose to do instead was remove the meta files and include the arguments in the Makefile using PNG2ASSETSFLAGS. This follows the existing convention already used for LCCFLAGS. Hopefully this is easier to parse for future users, and more maintainable. 

I did set the default behavior for GBC to noflip, as I believe that is the most user-friendly option and the LCCFLAGS for GBC are already set for gb and gbc compatibility (-Wm-yc), so this mirrors that.

I also added the ability to set the title of the ROM within the cartridge header, in addition to the existing option to name the file itself.

Please let me know if there's a better way to accomplish similar results. It didn't seem like the meta files had any other use, but I don't understand the history of why they currently exist, so I may be missing something. I did confirm that the build still works with these modifications, and the resulting ROMs work on emulators.

### DIFF
--- a/gbdk-lib/examples/cross-platform/logo/Makefile
+++ b/gbdk-lib/examples/cross-platform/logo/Makefile
@@ -12,6 +12,13 @@ PNG2ASSETS = $(GBDK_HOME)bin/png2asset
 # Possible are: gb gbc pocket megaduck sms gg
 TARGETS=gb gbc pocket megaduck sms gg nes
 
+# You can set the name of the resulting ROM file here
+PROJECTNAME = logo
+
+# You can set the name of the ROM title embedded in the ROM header here
+TITLE = LOGO # must be all upper case, maximum 15 characters
+
+
 # Configure platform specific LCC flags here:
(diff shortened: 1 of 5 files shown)


untoxa:
the idea is to keep *.meta files instead of hardcoding settings into the makefile itself, because projects usually have a lot of graphics resources and each file may have its own settings.

brettinman:
That makes sense, thank you for the context. I'll think of a better way to achieve this.

bbbbbr:
> That makes sense, thank you for the context. I'll think of a better way to achieve this.

Another way to proceed would be adding some of this information to the Readme.md for the example.  That can help instruct the users on why it is designed this way and how to make changes.

PR #627, opened 2024-02-23
Platformer template

LaroldsJubilantJunkyard:
**cross-platform Mario-style 2d platformer w/ 3 levels.**

Levels are generated from png2asset. Two of the 3 levels share the same tileset to show people how shared tilesets work. the third has it's own tileset.

**Features:**

- smooth motion
- mario-style jumping (hold a for higher jump)
- turning animations
- tilemap collision
- multiple levels


michel-iwaniec:
Good work!

However, latest update (b0890aaf72be51f35c9bbcee983046003448eb0d) gives me build errors like these for SMS:

../../../bin/lcc  -Wl-j -Wm-yoA -Wm-ya4 -Wb-ext=.rel -Wb- -Wm-ys -Wl-yt0x1B -autobank  -mz80:sms -Igen/sms/src -Ires -c -o obj/sms/common.o src/common.c
src/common.h:9: warning 283: function declarator with no prototype
src/common.c:9: warning 283: function declarator with no prototype
src/common.c:70: error 171: not allowed in a critical section or critical function
make[2]: *** [Makefile:95: obj/sms/common.o] Error 1
make[2]: Leaving directory '/c/geeky/gbdk-2020-mainline/build/gbdk/examples/cross-platform/platformer_template'
make[1]: *** [Makefile.targets:44: sms] Error 2
make[1]: Leaving directory '/c/geeky/gbdk-2020-mainline/build/gbdk/examples/cross-platform/platformer_template'
make: *** [Makefile:14: platformer_template/.] Error 2

However, after commenting out the new "CRITICAL" sections I can build it.

Was there a specific reason these were added?

After commenting these out I can build everything again. 

But I noticed these issues:
1) When jumping past the top of the screen, there is a strange trace of the player left behind for a number of frames. I suspect this has to do with how the shadow OAM is / isn't cleared?

2) Optional to fix: The player still gets stuck into the floor sometimes, requires jumping to get unstuck. But I don't necessarily think this bug should block merging as it could be fixed later, and only locks things up very temporarily. Needs an investigation at some point though for a more polished experience.

3) Optional to fix: It would be nice if the screen was off when loading the level tile / tilemap, as the visible tile changes are a bit distracting. It should also make level loading faster.

bbbbbr:
This can wait until after the other code changes, but it would be nice to not include the entries in the .gitignore:
> gbdk-lib/examples/cross-platform/platformer_template/obj/
> gbdk-lib/examples/cross-platform/platformer_template/dist/
> gbdk-lib/examples/cross-platform/platformer_template/gen/

It can be a little more work when iterating, but I recommend only building examples out-of-tree. That way it also closely matches what the user will experience when they build the examples, and may help catch path related issues.

bbbbbr:
A couple suggestions:

**Constants**
Values describing hardware characteristics, such as the 224 (and later 240) in:
(src/camera.c) 
> `#define WRAP_SCROLL_Y(y) ((y) % 224u)`
Should be derived from the DEVICE_* constants if possible.

**Current bank var**
We are gradually migrating from `_current_bank` to `CURRENT_BANK`

**Comments**
Would be nice to see brief comments on:
-  update_column_left(uint8_t map_pos_x) - What it does, why there is different handling
-  update_column_right(uint8_t map_pos_x) - What it does
-  setBKGPalettes(), SetPlayerPalettes() - Explaining about the purpose of the VDP writes
- showCentered() - Purpose of the function, why DISPLAY_OFF/ON are wrapped in a critical section
-  IsTileSolid(() - Where the 250 in `if(row>250||column>=currentLevelWidthInTiles)` comes from and why that's an early exit

LaroldsJubilantJunkyard:
**For the update column functions**

> Would be nice to see brief comments on:
> 
> * update_column_left(uint8_t map_pos_x) - What it does, why there is different handling
> * update_column_right(uint8_t map_pos_x) - What it doe


i got these from the large map example. there wasn't much comments there either. i can't say really, i can see that it does something different in `update_column_left` depending on the device screen buffer width.

**_`setBKGPalettes(), SetPlayerPalettes() - Explaining about the purpose of the VDP writes`_**

I used what i got from some other other examples

_**`IsTileSolid(() - Where the 250 in if(row>250||column>=currentLevelWidthInTiles) comes from and why that's an early exit`**_

i've changed that to be a variable. It's purpose is to check for wrap around for unsigned integers.

**The 'ShowCentenred' function was questioned, regarding it's purpose**

The different platforms have different screen sizes. If it were a game boy only game i could just use a 160x144 graphic. but that won't be centered for other platforms. So i created a function that would determine the difference and position the graphic in the center

**As for the gitignore changes for the generated folders:**

> This can wait until after the other code changes are settled, but it would be nice to not include the entries in the .gitignore:
> 
> > gbdk-lib/examples/cross-platform/platformer_template/obj/
> > gbdk-lib/examples/cross-platform/platformer_template/dist/
> > gbdk-lib/examples/cross-platform/platformer_template/gen/
> 
> It can be a little more work when iterating, but I recommend only building examples out-of-tree. That way it also closely matches what the user will experience when they build the examples, and may help catch path related issues.

that has been removed

gbdk-lib/examples/cross-platform/platformer_template/src/player.c
```diff
@@ -0,0 +1,355 @@
...
+
+uint8_t playerJumpIncrease = 0;
+uint8_t threeFrameCounter=0;
+uint16_t playerX, playerY;
+int16_t playerXVelocity, playerYVelocity;
+
+    // Flip horizontally, if we aren't facing right
+
+#if defined(NINTENDO_NES)
+    const uint8_t baseProp=4;
```
michel-iwaniec, line 41: I don't think this baseProp parameter is needed here, as it probably doesn't do what you want anyway.

The baseProp parameter to move_metasprite_ex will be added to the properties of the metasprite. However, adding 4 to it seems kind of pointless as that represents bit 2, and that bit doesn't even exist in the PPUs OAM memory. (is always zero after OAM DMA)

Furthermore, as you already have the different directions for the player as separate .png files, flipping the player here is redundant, even if you used the correct bit.

gbdk-lib/examples/cross-platform/platformer_template/src/player.c
```diff
@@ -0,0 +1,355 @@
...
+    
+    // Set up color palettes
+    #if defined(SEGA)
+            set_sprite_palette(baseProp, 1, PLAYER_PALETTES);
+    #elif defined(GAMEBOY)
+        if (_cpu == CGB_TYPE) {
+            set_sprite_palette(OAMF_CGB_PAL0, 1, PLAYER_PALETTES);
+        }
+    #elif defined(NINTENDO_NES)
+        set_sprite_palette(baseProp, 4, PLAYER_PALETTES);
```
michel-iwaniec, line 105: I don't think using baseProp here is very useful, as it has nothing to do with the flipping described earlier.

As you already have separate if-clauses for each system, I would just have the first parameter be the correct value there. It will make things more readable.

As a side-note, I think we'll need to have a review/discussion about how to handle the palette indices for CGB vs NES. They are a little bit inconsistent at the moment. (but that can be done after merging this)

gbdk-lib/examples/cross-platform/platformer_template/src/main.c
```diff
@@ -0,0 +1,87 @@
+#include <gbdk/platform.h>
+#include <stdint.h>
+#include "player.h"
+#include "common.h"
+#include "level.h"
+#include "Camera.h"
```
bbbbbr, line 6: Capitalization here causes a case sensitivity error on Linux. Should be "camera.h"
src/main.c:6:10: fatal error: Camera.h: No such file or directory
    6 | #include "Camera.h"

gbdk-lib/examples/cross-platform/platformer_template/Makefile
```diff
@@ -0,0 +1,129 @@
...
+GENDIR      = gen/$(EXT)/src
+BINDIR      = build/$(EXT)
+MKDIRS      = $(GENDIR) $(OBJDIR) $(BINDIR) # See bottom of Makefile for directory auto-creation
+BINS	    = $(OBJDIR)/$(PROJECTNAME).$(EXT)
+CSOURCES    = $(foreach dir,$(SRCDIR),$(notdir $(wildcard $(dir)/*.c))) 
+GENSOURCES    = $(foreach dir,$(GENDIR), $(wildcard $(dir)/*.c))
+ASMSOURCES  = $(foreach dir,$(SRCDIR),$(notdir $(wildcard $(dir)/*.s)))
+OBJS       = $(CSOURCES:%.c=$(OBJDIR)/%.o) $(ASMSOURCES:%.s=$(OBJDIR)/%.o)
+
+LCCFLAGS += -Wl-j -Wm-yoA -Wm-ya4 -Wb-ext=.rel -Wb- $(LCCFLAGS_$(EXT)) # This adds the current platform specific LCC Flags
```
bbbbbr, line 41: There is a stray `-Wb- ` at the end causing bankpack to issue a warning due to the detached ` - `. Probably just remove it.

"../../../"bin/bankpack -plat=nes -mapper=30 -ext=.rel - -yt0x1B -lkout=/tmp/lcc2393229.lk obj/nes/camera.o obj/nes/common.o obj/nes/level.o obj/nes/main.o obj/nes/player.o /tmp/lcc2393220.o /tmp/lcc2393221.o /tmp/lcc2393222.o /tmp/lcc2393223.o /tmp/lcc2393224.o /tmp/lcc2393225.o /tmp/lcc2393226.o /tmp/lcc2393227.o /tmp/lcc2393228.o
BankPack: Warning: Ignoring unknown option -

bbbbbr:
Still seeing a bunch of warnings related to missing `void` in function definitions and declarations with no parameters. Would like to see all those tidied up.

Ires -I"../../../"include -c src/camera.c -o obj/gb/camera.o
src/level.h:19: warning 283: function declarator with no prototype
src/camera.c:56: warning 283: function declarator with no prototype

../../../bin/lcc $Id: lcc.c,v 2.0 2024/04/26 20:23:19 gbdk-2020 Exp $
"../../../"bin/sdcc -msm83 --no-std-crt0 --fsigned-char --use-stdout -D__PORT_sm83 -D__TARGET_gb -Wa-pogn -I"../../../"lib/gb -Igen/gb/src -Ires -I"../../../"include -c src/common.c -o obj/gb/common.o
src/common.h:9: warning 283: function declarator with no prototype
src/common.c:9: warning 283: function declarator with no prototype

../../../bin/lcc $Id: lcc.c,v 2.0 2024/04/26 20:23:19 gbdk-2020 Exp $
"../../../"bin/sdcc -msm83 --no-std-crt0 --fsigned-char --use-stdout -D__PORT_sm83 -D__TARGET_gb -Wa-pogn -I"../../../"lib/gb -Igen/gb/src -Ires -I"../../../"include -c src/level.c -o obj/gb/level.o
src/common.h:9: warning 283: function declarator with no prototype
src/level.c:72: warning 283: function declarator with no prototype

../../../bin/lcc $Id: lcc.c,v 2.0 2024/04/26 20:23:19 gbdk-2020 Exp $
"../../../"bin/sdcc -msm83 --no-std-crt0 --fsigned-char --use-stdout -D__PORT_sm83 -D__TARGET_gb -Wa-pogn -I"../../../"lib/gb -Igen/gb/src -Ires -I"../../../"include -c src/main.c -o obj/gb/main.o
src/player.h:8: warning 283: function declarator with no prototype
src/player.h:9: warning 283: function declarator with no prototype
src/common.h:9: warning 283: function declarator with no prototype
src/level.h:19: warning 283: function declarator with no prototype
src/camera.h:10: warning 283: function declarator with no prototype

../../../bin/lcc $Id: lcc.c,v 2.0 2024/04/26 20:23:19 gbdk-2020 Exp $
"../../../"bin/sdcc -msm83 --no-std-crt0 --fsigned-char --use-stdout -D__PORT_sm83 -D__TARGET_gb -Wa-pogn -I"../../../"lib/gb -Igen/gb/src -Ires -I"../../../"include -c src/player.c -o obj/gb/player.o
src/common.h:9: warning 283: function declarator with no prototype
src/camera.h:10: warning 283: function declarator with no prototype
src/level.h:19: warning 283: function declarator with no prototype
src/player.c:75: warning 283: function declarator with no prototype
src/player.c:92: warning 283: function declarator with no prototype
src/player.c:112: warning 283: function declarator with no prototype
src/player.c:154: warning 283: function declarator with no prototype

bbbbbr:
Unless the data pointed to by `currentLevelMap` gets modified, it can be `const uint8_t * ` instead of `uint8_t *` which will clear up the warnings below.

Ires -I"../../../"include -c src/level.c -o obj/gb/level.o
src/level.c:110: warning 196: pointer target lost const qualifier
src/level.c:135: warning 196: pointer target lost const qualifier
src/level.c:158: warning 196: pointer target lost const qualifier

bbbbbr:
The aliasing on the big text labels is kind of rough, attached are touchups for those on two of the files
paltformer_template_font_2
platformer_template_font_1
title-screen
next-level

bbbbbr:
Something strange is happening when I tune the player movement to be a little faster and less floaty. Looks like the player speed may never quite reach zero... a bit of moon walking. :)

#define GRAVTY 45
#define PLAYER_CHARACTER_JUMP_VELOCITY 550
#define PLAYER_CHARACTER_WALK_VELOCITY 325
#define PLAYER_CHARACTER_RUN_VELOCITY 425

https://github.com/gbdk-2020/gbdk-2020/assets/8322401/aa5cf2ed-8bdd-4dd8-ba31-9c6bf35d6c54

gbdk-lib/examples/cross-platform/platformer_template/src/player.c
```diff
@@ -0,0 +1,359 @@
...
+            }
+            
+        }
+        
+    }else{
+
+        // Move the x velocity towards 0
+        if(playerXVelocity>=GROUND_FRICTION)playerXVelocity-=GROUND_FRICTION;
+        else if(playerXVelocity<=GROUND_FRICTION)playerXVelocity+=GROUND_FRICTION;
+        else playerXVelocity=0;
```
bbbbbr, line 216: Seems like making this change resolves the never reaches zero movement issue (when using different motion settings). Might be another way to fix it too.
```
        // Move the x velocity towards 0
        if (playerXVelocity > 0) {
          if (playerXVelocity >= GROUND_FRICTION) playerXVelocity -= GROUND_FRICTION;
          else playerXVelocity=0;
        }
        else if (playerXVelocity < 0) {
            if (playerXVelocity <= GROUND_FRICTION) playerXVelocity += GROUND_FRICTION;
            else playerXVelocity=0;
        }
```

bbbbbr:
Seeing a couple frame glitch only when running right and then pressing and holding down left to turn and run left. 

https://github.com/gbdk-2020/gbdk-2020/assets/8322401/b1901786-908d-4979-9126-19630bb66952

bbbbbr:
If other's don't have a preference I'd like to suggest these motion settings instead. They may be a little closer to typical game physics (though still a bit softer maybe).

```
#define GRAVTY 45
#define PLAYER_CHARACTER_JUMP_VELOCITY 550
#define PLAYER_CHARACTER_WALK_VELOCITY 325
#define PLAYER_CHARACTER_RUN_VELOCITY 425
```

PR #630, opened 2024-03-06
fix mode flags such as M_NO_SCROLL being discarded when setting M_TEXT_OUT

tihmstar:
Setting `mode(M_TEXT_OUT);` causes the `.tmode_out` routine to be called internally.
That routine always sets `.mode` to `M_TEXT_OUT`, which discards additional flags.
This commit, makes sure that flags like `M_NO_SCROLL` persist this routine and disabling scrolling the screen works as intended


untoxa:
Michael, can you please review this? I think NES port has different default behavior...

PR #670, opened 2024-05-16
* Change platformer template to blank screen during all of level loading

michel-iwaniec:


LaroldsJubilantJunkyard (approved):
This sounds about right. If things work quicker, that's great. There was no intentional reason for the slower screen filling.

PR #675, opened 2024-05-23
text_scroller example: Fix problem with '*' fill characters not appearing on GG/SMS

michel-iwaniec:
text_scroller example: Fix problem with '*' fill characters not appearing on GG/SMS
* Move fill_bkg_rect to happen *after* the first printf, as the printf causes a font load which causes a screen clear
* Make fill_bkg_rect fill start at second row, to not overwrite the first printf


untoxa (commented):
Already fixed

PR #679, opened 2024-05-25
Mention that HIDE_BKG/SHOW_BKG don't work on CGB

Tempestro:
Adds a mention to the docs that HIDE_BKG and SHOW_BKG do not work while running in CGB mode, the relevant bit instead overrides all background/window priority over the sprite layer, see Pandocs 4.4

### DIFF
--- a/gbdk-lib/include/gb/gb.h
+++ b/gbdk-lib/include/gb/gb.h
@@ -893,8 +893,8 @@ void hiramcpy(uint8_t dst, const void *src, uint8_t n) OLDCALL PRESERVES_REGS(b,
 
     Doesn't work in CGB mode - the bit is reused to control sprite priority
     over background and window layers instead.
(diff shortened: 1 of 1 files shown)


bbbbbr:
Good find. The mismatch in behavior and docs for GBC is probably due to GBDK being released a few years (!) before the GBC hardware came out, and then the docs never got revised to accommodate that.

To keep doxygen from merging each "If..." sentence into sentences above it a leading `\li ` should be added in front of each one. That will also convert them into bullet points in the HTML and PDF docs output.

For example: 
https://github.com/gbdk-2020/gbdk-2020/blame/1841b0a68921619e13205326ac6ae262296f79ca/gbdk-lib/include/gb/gb.h#L1198

bbbbbr:
Thanks!

PR #681, opened 2024-05-30
bankpack: Update to use parse Defs as 32-bit numbers instead of 24bit

michel-iwaniec:


bbbbbr:
Expanded to support both in #682

PR #698, opened 2024-06-21
Palette Generation broken when using -source_tileset option

fingerflinger:
If the source tileset includes duplicate palette information, the exporter will exclude the redundant palettes. However, the exporter was checking if i != 0 to check for the first time through the loop, and this is not true in the case of the source_tileset excluding redundant palettes. This was generating a leading comma in the output, which causes a syntax error.

### DIFF
--- a/gbdk-support/png2asset/export_c_file.cpp
+++ b/gbdk-support/png2asset/export_c_file.cpp
@@ -113,9 +113,10 @@ static void export_c_palette_data(PNG2AssetData* assetData, FILE* file) {
 
     if (!(assetData->args->use_structs) || (assetData->args->includeTileData)) {
         fprintf(file, "const palette_color_t %s_palettes[%d] = {\n", assetData->args->data_name.c_str(), (unsigned int)exportOpt.color_count);
-        for(size_t i = exportOpt.color_start / assetData->image.colors_per_pal; i < assetData->image.total_color_count / assetData->image.colors_per_pal; ++i)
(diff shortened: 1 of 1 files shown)


gbdk-support/png2asset/export_c_file.cpp
```diff
@@ -113,9 +113,10 @@ static void export_c_palette_data(PNG2AssetData* assetData, FILE* file) {
 
     if (!(assetData->args->use_structs) || (assetData->args->includeTileData)) {
         fprintf(file, "const palette_color_t %s_palettes[%d] = {\n", assetData->args->data_name.c_str(), (unsigned int)exportOpt.color_count);
-        for(size_t i = exportOpt.color_start / assetData->image.colors_per_pal; i < assetData->image.total_color_count / assetData->image.colors_per_pal; ++i)
+        const size_t i0 = exportOpt.color_start / assetData->image.colors_per_pal;
```
bbbbbr, line 116: Instead of `i0` this should be named something such as `palette_start`.

And along those lines, for improved readability of the original code this could be added as well:
`const size_t total_palette_count = assetData->image.total_color_count / assetData->image.colors_per_pal;`

Making the loop easier to visually parse:
`for (size_t i = palette_start; i < total_palette_count; ++i)`

bbbbbr:
Thanks for opening the PR to fix this. Aside from the above requested change this looks good.

fingerflinger:
Sounds good, updated. Thanks for the review

bbbbbr:
Great, thanks!

PR #712, opened 2024-08-21
Feature/text  & dialogue example

LaroldsJubilantJunkyard:
Adding cross-platform examples for drawing text and dialogue.

- Drawing large amounts of text
- Waiting for input
- Starting new lines to avoid cutting off words
- Typewriter Effect


gbdk-lib/examples/cross-platform/text_basic/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+    else if(character>='0'&&character<='9')vramTile = (character-'0')+27;
+    else if(character=='!')vramTile = 37;
+    else if(character==':')vramTile = 38;
+    else if(character=='?')vramTile = 39;
+    else if(character=='/')vramTile = 40;
+    else if(character=='=')vramTile = 41;
+    else if(character==',')vramTile = 42;
+    else if(character=='.')vramTile = 43;
+    else if(character=='<')vramTile = 44;
+    else if(character=='>')vramTile = 45;
```
bbbbbr, line 25: (This applies for all of the examples with the same code)

- It's a little smaller code wise and perhaps a bit more efficient to use a case/select for the misc tiles. Perhaps more so as font characters are added. At some point a recode table might become more efficient, but this is probably good to start with.
- Having spaces in the if statements and the code they execute, and aligning the formatting makes it a bit easier to read and understand them.
```C
    if      (character >= 'a' && character <= 'z') vramTile = (character - 'a') + 1;
    else if (character >= 'A' && character <= 'Z') vramTile = (character - 'A') + 1;
    else if (character >= '0' && character <= '9') vramTile = (character - '0') + 27;
    else {
        switch(character) {
            case '!': vramTile = 37; break;
            case ':': vramTile = 38; break;
            case '?': vramTile = 39; break;
            case '/': vramTile = 40; break;
            case '=': vramTile = 41; break;
            case ',': vramTile = 42; break;
            case '.': vramTile = 43; break;
            case '<': vramTile = 44; break;
            case '>': vramTile = 45; break;
        }
    }
```

gbdk-lib/examples/cross-platform/text_basic/src/main.c
```diff
@@ -0,0 +1,76 @@
...
+        index++;
+
+    }
+}
+
+void main(void)
+{
+    SHOW_BKG;
+
+    #if !defined(SEGA) && !defined(NINTENDO_NES)
```
bbbbbr, line 62: For the basic example which doesn't have sound effects it's better to keep it simple and omit the GB sound init code.

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        if(typewriterDelay>0){
+
+            #if !defined(SEGA) && !defined(NINTENDO_NES)
+            
+            // Play a basic sound effect
+            NR10_REG = 0x34;
+            NR11_REG = 0x81;
+            NR12_REG = 0x41;
```
bbbbbr, line 73: For the MegaDuck values written to NR12 need a Nybble swap (true for other channels as well, but not applicable here). There's an example of this in `examples/gb/sound`

```C
inline uint8_t translate_envelope(uint8_t value) {
	#ifdef MEGADUCK
        // Mega Duck has nybbles swapped for NR12, NR22, NR42 audio envelope registers
		return ((uint8_t)(value << 4) | (uint8_t)(value >> 4));
	#else
		return value;
	#endif
}
```
Then
```C
NR12_REG = translate_envelope(0x41);
```

https://github.com/gbdk-2020/gbdk-2020/blob/c97d278cd4178dbe0bd3d66589821d7f6f31e711/gbdk-lib/examples/gb/sound/sound.c#L7C1-L14C2

MegaDuck sound info in the GBDK docs:
https://gbdk-2020.github.io/gbdk-2020/docs/api/docs_supported_consoles.html#autotoc_md190

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
```diff
@@ -0,0 +1,108 @@
...
+            
+            // reset for the next row
+            vramAddress = get_bkg_xy_addr(column,++row);
+            
+            columnIndex=0;
+        }
+
+        if(typewriterDelay>0){
+
+            #if !defined(SEGA) && !defined(NINTENDO_NES)
```
bbbbbr, line 68: I'll leave the details to @untoxa and @michel-iwaniec since they're more familiar with the platforms, but it would be nice to have sound for SMS/GG and NES as well. it could be split out to a function to keep it manageable.

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
```diff
@@ -0,0 +1,108 @@
...
+            NR10_REG = 0x34;
+            NR11_REG = 0x81;
+            NR12_REG = 0x41;
+            NR13_REG = 0x7F;
+            NR14_REG = 0x86;
+
+            #endif
+
+            // Wait some frames
+            // This creats a typewriter effect
```
bbbbbr, line 80: Typo: `creates`

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
```diff
@@ -0,0 +1,108 @@
...
+    NR51_REG = 0xFF; // Turn on sound fully
+
+    #endif
+
+    set_native_tile_data(0,Font_TILE_COUNT,Font_tiles);
+
+    fill_bkg_rect(0,0,DEVICE_SCREEN_WIDTH,DEVICE_SCREEN_HEIGHT,0);
+
+    // By passing 3 as the final argument, the game boy will wait 3 frames between each character
+    DrawText("This is an how you draw text on the screen in GBDK. The code will automatically jump to a new line, when it reaches the end of the row.",3);
```
bbbbbr, line 107: Fix typo + some rephrasing (since it's one of many possible ways, not the only way): -> `This is a way to draw text on the screen in GBDK.` (second sentence looks good)

The `3` would be easier for people to quickly grasp it's purpose when looking at the function call if it was a named constant, such as `3_FRAMES`.

gbdk-lib/examples/cross-platform/text_advanced_dialogue/src/main.c
```diff
@@ -0,0 +1,374 @@
...
+        windowYPosition = (DEVICE_SCREEN_HEIGHT << 3)<<3;
+        MoveWindow();
+
+    #endif
+    
+    while(TRUE){
+
+        // We'll pass in one long string, but the game will present to the player multiple pages.
+        // By passing 3 as the final argument, the game boy will wait 3 frames between each character
+        DrawTextAdvanced("This is an how you draw text on the screen in GBDK. The code will automatically jump to a new line, when it cannot fully draw a word.  When you reach three lines, it will wait until you press A. After that, it will start a new page and continue. The code will also automatically start a new page after periods. For every page, the code will dynamically populate VRAM. Only letters and characters used, will be loaded into VRAM.");
```
bbbbbr, line 371: It would be helpful to put the "Press A button" info at the very beginning so it's clear that they should do that to continue displaying more text.

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
untoxa: Yes, of course. If you add sound to the cross-platform example, better make it work on all platforms. It it is not essential, then it is better to exclude that.

bbbbbr:
Is the one line gap between every line of text intentional? The rendered text says `When you reach three lines, it will wait until you press A.`, but the first part gets scroll clipped and only ends up showing the last 2 of those lines in the demo.

See attached video

I guess regardless of whether it is intentional or not, to give the example more versatility it should be handled programmatically.
Some alternative suggestions:
- A flag for the print function for how many lines to advance when wrapping to a new line
- A `\n` line break character in the string and a handler in the printing. That way people can easily specify gaps when they want them

https://github.com/user-attachments/assets/95c91c37-71d4-4433-9025-2e7089579955

LaroldsJubilantJunkyard:
> Is the one line gap between every line of text intentional? The rendered text says `When you reach three lines, it will wait until you press A.`, but the first part gets scroll clipped and only ends up showing the last 2 of those lines in the demo.
> 
> See attached video
> 
> I guess regardless of whether it is intentional or not, to give the example more versatility it should be handled programmatically. Some alternative suggestions:
> 
> * A flag for the print function for how many lines to advance when wrapping to a new line
> * A `\n` line break character in the string and a handler in the printing. That way people can easily specify gaps when they want them
> 
>  text_advanced_line_gaps.mp4

Yes, it's meant to be that way. Mimicking The Legend of Zela: Links Awakening

image

Two rows of text. The second  row slides up to the middle, then the top. A new second row is created when the next part of the text. I meant to update the text. This is what i think i should change it to:

> When the code reaches a period or question mark, it will wait for you to press A. Afterwards, It will start a new page and continue.  The code will automatically jump to a new line, when it cannot fully draw a word.  When both rows of text are full, the code will slide the current text upwards and continue. For every page, the code will dynamically populate VRAM. Only letters and characters used, will be loaded into VRAM.

As for  adding a line break.... That wasn't in my original vision. You guys can add that later if you want.

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
LaroldsJubilantJunkyard: I don't know how to do sound for those platforms. I also don't see any examples on it in their folders.

bbbbbr: There is some code in the port of Black Castle to GG/SMS and NES that I think attempts to vaguely emulate a GB sound interface for the other platforms.  @untoxa and @michel-iwaniec have a better understanding of the code there and whether it's appropriate to use. It may be that it's too special-cased to that game and they don't want users to take it as something to mimic.
https://github.com/untoxa/BlackCastle/blob/74dc83cb56df28182cdd0f2ae5019828e440f931/src/sfx.c

The alternative is to follow Toxa's suggestion and omit the sound feature since it's not essential.

bbbbbr:
Got it that it's by design.

Since this is intended as an example users may build on or at least use as reference code, it seems like it should have that flexibility to not have the spaces from the start. We can expect some portion of users will want to have all 3 lines of text in there.

If controlled by a variable it seems like it should not be too hard for the example to either advance the line once or twice. And that variable can be commented to explain how it's used.

gbdk-lib/examples/cross-platform/text_typewriter/src/main.c
LaroldsJubilantJunkyard: > There is some code in the port of Black Castle to GG/SMS and NES that I think attempts to vaguely emulate a GB sound interface for the other platforms. @untoxa and @michel-iwaniec have a better understanding of the code there and whether it's appropriate to use. It may be that it's too special-cased to that game and they don't want users to take it as something to mimic. https://github.com/untoxa/BlackCastle/blob/74dc83cb56df28182cdd0f2ae5019828e440f931/src/sfx.c
> 
> The alternative is to follow Toxa's suggestion and omit the sound feature since it's not essential.

audio removed from the typewriter and advanced dialogue examples

bbbbbr:
Aside from a issue with the window on the GB the changes look good so far.

For the GB: it looks like the window is supposed to scroll up from the bottom but instead it shows up at the top immediately. Might be a couple things:
- Need to set the initial position to below the bottom of the viewport
- `desiredWindowPosition` is in pixels but it's tested against tile coordinates `(windowYPosition>>3)`
- Likewise, `move_win()` should receive coordinates in pixels
- Since it's GB only and the range is 0-144, it can be a `uint8_t`
- Reduced the speed a little

This should resolve it. I had trouble adding it into the PR.
```Patch


gbdk-lib/examples/cross-platform/text_advanced_dialogue/src/main.c:30

-int16_t windowYPosition=0;
+uint8_t windowYPosition= DEVICE_SCREEN_PX_HEIGHT; // Start window below bottom of screen

 void MoveWindow(void){
     #if !defined(SEGA) &&  !defined(NINTENDO_NES)
 
-    int16_t y = windowYPosition>>3;
+    uint8_t y = windowYPosition;
 
     move_win(7,y);
     #endif
 }
 
 void ShowDialgoueBox(void){
 
     ClearDialogueBox();
 
-    int16_t desiredWindowPosition = (DEVICE_SCREEN_HEIGHT<<3)-(DIALOG_BOX_HEIGHT*8);
+    uint8_t desiredWindowPosition = DEVICE_SCREEN_PX_HEIGHT - (DIALOG_BOX_HEIGHT*8);
 
-    while((windowYPosition>>3)>desiredWindowPosition){
+    while(windowYPosition > desiredWindowPosition){
 
-        windowYPosition-=10;
+        windowYPosition-=4;
         MoveWindow();
         vsync();
     }
 }
 
 void HideDialgoueBox(void){
 
-    int16_t desiredWindowPosition = (DEVICE_SCREEN_HEIGHT<<3);
+    uint8_t desiredWindowPosition = DEVICE_SCREEN_PX_HEIGHT;
 
-    while((windowYPosition>>3)<desiredWindowPosition){
+    while (windowYPosition < desiredWindowPosition) {
 
-        windowYPosition+=10;
+        windowYPosition+=4;
         MoveWindow();
         vsync();
     }
 }

```

LaroldsJubilantJunkyard:
> Aside from a issue with the window on the GB the changes look good so far.
> 
> For the GB: it looks like the window is supposed to scroll up and display near the bottom but instead it shows up at the very top immediately. Might be a couple things:
> 
> * Need to set the initial position to below the bottom of the viewport
> * `desiredWindowPosition` is in pixels but it's tested against tile coordinates `(windowYPosition>>3)`
> * Likewise, `move_win()` should receive coordinates in pixels
> * Since it's GB only and the range is 0-144, it can be a `uint8_t`
> * Reduced the speed a little
> 
> This should resolve it. I had trouble adding it into the PR.
> 
> ```diff
> gbdk-lib/examples/cross-platform/text_advanced_dialogue/src/main.c:30
> 
> -int16_t windowYPosition=0;
> +uint8_t windowYPosition= DEVICE_SCREEN_PX_HEIGHT; // Start window below bottom of screen
> 
>  void MoveWindow(void){
>      #if !defined(SEGA) &&  !defined(NINTENDO_NES)
>  
> -    int16_t y = windowYPosition>>3;
> +    uint8_t y = windowYPosition;
>  
>      move_win(7,y);
>      #endif
>  }
>  
>  void ShowDialgoueBox(void){
>  
>      ClearDialogueBox();
>  
> -    int16_t desiredWindowPosition = (DEVICE_SCREEN_HEIGHT<<3)-(DIALOG_BOX_HEIGHT*8);
> +    uint8_t desiredWindowPosition = DEVICE_SCREEN_PX_HEIGHT - (DIALOG_BOX_HEIGHT*8);
>  
> -    while((windowYPosition>>3)>desiredWindowPosition){
> +    while(windowYPosition > desiredWindowPosition){
>  
> -        windowYPosition-=10;
> +        windowYPosition-=4;
>          MoveWindow();
>          vsync();
>      }
>  }
>  
>  void HideDialgoueBox(void){
>  
> -    int16_t desiredWindowPosition = (DEVICE_SCREEN_HEIGHT<<3);
> +    uint8_t desiredWindowPosition = DEVICE_SCREEN_PX_HEIGHT;
>  
> -    while((windowYPosition>>3)<desiredWindowPosition){
> +    while (windowYPosition < desiredWindowPosition) {
>  
> -        windowYPosition+=10;
> +        windowYPosition+=4;
>          MoveWindow();
>          vsync();
>      }
>  }
> ```

all resolved! I also made the typewriter delay, and line skip delay configurable.

PR #752, opened 2025-03-21
Adding support for layered metasprites

LaroldsJubilantJunkyard:
Developers will be able to pass "-layered-sprites" when generating metasprites. This will allow multi-palette sprites to properly be converted and shown.

When enabled, the option should show a warning. So that developers are aware of the sprite/scanline limits.

### DIFF
--- a/gbdk-support/png2asset/metasprites.cpp
+++ b/gbdk-support/png2asset/metasprites.cpp
@@ -18,6 +18,58 @@
 
 using namespace std;
 
+void ProcessMetaspriteTile(int _x, int _y, int x, int y, int* last_x, int* last_y, int _w, int _h, int pivot_x, int pivot_y, PNG2AssetData* assetData, int palleteIndex, MetaSprite& mt_sprite) {
+    Tile tile(assetData->image.tile_h * assetData->image.tile_w);
+
+    // For sprites, unlike maps, Tiles are only kept and processed if NOT Empty
+    // This default behavior can be overridden with -spr_no_optimize (which sets keep_empty_sprite_tiles and keep_duplicate_tiles to TRUE)
+    bool tile_not_empty = (assetData->image.ExtractTile(x, y, tile, assetData->args->sprite_mode, assetData->args->export_as_map, assetData->args->use_map_attributes, assetData->args->bpp, palleteIndex));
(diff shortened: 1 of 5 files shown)


bbbbbr:
(copying comments over from chat)

I like the idea of this feature, but have a couple thoughts:

First- I think this is a cool feature and might be nice to have. I've done layered sprites in a gbdk program before and it was tedious to process they layers separately, so I know the benefit this would bring.

Second- While it's rather gradual, we are moving toward a new release. Changes to the support tools settled down a few months ago and they've now had a decent amount of testing/exercise and bug fixing. So we can be reasonably confident they're in good shape.
 
My preference would be to keep them in that more stable state and defer sizable feature adds/code changes until after the release. Particularly for png2asset which has the most complex and often tricky behavior of all the support tools.

bbbbbr:
Third: 

Implementation- This explanation from the chat discussion is helpful:
> *when -1, everything stays the same. otherwise, it will use that for the tile's pallete. in addition, pixels outside of that palette are  changed to 0. (there's a thought in my head that i might need to adjust that, but i haven't articulated it yet)*

But it's hard to grasp that when reading through the changed code and trying to reconstruct what is going on and how it works. 

Given the png2asset codebase, I'm not sure if there is a better more graceful way to do it - but I'd love to be able to read and maintain code for this feature where it was immediately intuitive and clear what it was doing and how it was doing it when creating the multiple layers for sprite tiles.

Something where it would be very code-readable like: 
```C
... GetMetaSpriteLayers(...) {
  // When support for layered sprites is enabled, GetMetaSprite() is
  // called repeatedly on the same location. On each pass it <does stuff>
  // to extract a single tile as a layer with as many pixels possible given a
  // console's palette limitations. Once no more sprite pixel data is found
  // it will return false indicating completion (or exhaustion of palettes).
  while (GetMetaSprite(...));
}
```

paletteIndex - Would also love to see if there was a way to do that a little more cleanly without passing the paletteIndex var down through all those layers, esp when we've made some gains during the big refactoring in reducing the function arg count. Sometimes it's needed of course, but would love if there was a more streamlined option.

Indexed/vs rgb image - There might be some ways to naively support rgb images and not require indexed palettes. The output may consume palettes more wastefully, but it could still be convenient in the cases where it doesn't matter.

PR #789, opened 2025-06-16
Rumble Pak example

EV3lindaboi:
I saw that GBDK does not have a Rumble Pak example. So I made one.
I thought of this because I added Rumble functionality to a game I am working. Had to look in the GBDK Discord server for that. So, the code is not fully mine. Credits for the Rumble functions themselves go to Toxa.


untoxa:
example seem very primitive, maybe there is no much sense in it. also there exist several different rumbles, not only the "MBC5 rumble". they all work similar, but on the different bits.

bbbbbr:
Thanks for opening the PR, good start.

1. I'd like to see the MBC/Rumble specific code moved to a separate header file similar to the [MBC3](https://github.com/gbdk-2020/gbdk-2020/tree/develop/gbdk-lib/examples/gb/mbc3_rtc/src) and MBC7 examples. You can also lift the file structure from those examples where they have a `src/` dir, etc.

2. It would also be useful to have a RUMBLE_ON() / RUMBLE_OFF() macro abstraction which might preserves the existing RAM bank (or a note that it doesn't).

3. In my experience using rumble in a game and running from a cart the default intensity was too high and somewhat distracting. So an addition to the example could be how to use partial duty cycles. (Edit: apparently also pokemon pinball adjusts the [rumble intensity](https://gbdev.io/pandocs/MBC5.html?highlight=rumble#rumble)).


For example (~roughly taken from canyon racer):
```C
#define RUMBLE_COUNT_MASK  0x07u
#define RUMBLE_COUNT_MAX  (RUMBLE_COUNT_MASK - 1u)
uint8_t rumble_intensity = 0u;
uint8_t rumble_count = 0u;

// Adjust intensity of Rumble
if (increase intensity) {
    if (rumble_intensity > 0) rumble_intensity--;
}
else if (decrease intensity) {
   if (rumble_intensity <  RUMBLE_COUNT_MAX) rumble_intensity++;
}
...
// Turn on rumbling for some duration based on an event
if (some event) {
    rumble_count = RUMBLE_COUNT_SOME_EVENT;
}
...
// Update cart rumble, this would be done once per frame, preferably in a function
if (rumble_count) {
    rumble_count--;

    if ((rumble_count & RUMBLE_COUNT_MASK) > rumble_intensity)
        MBC5_RUMBLE_ON;
    else
        MBC5_RUMBLE_OFF;
}
```

bbbbbr:
@untoxa wrote:
> also there exist several different rumbles, not only the "MBC5 rumble". they all work similar, but on the different bits.

- I've seen mention of MBC3 rumble in the GBStudio code but where can they be obtained?
- It doesn't appear to have an official Nintendo MBC number, is it an aftermarket mod to MBC3 carts?
- Are there emulators that support MBC3 rumble?

untoxa:
iirc, this MBC5 compatible cart has 32kb of RAM, and rumble bit is bit7 or something like that: https://shop.insidegadgets.com/product/gameboy-4mb-32kb-fram-flash-cart-ultra-low-power/

bbbbbr:
That MBC5 IG rumble cart is the one I have. But I was wondering if there are any others available on the market, and in particular the MBC3 one that GBStudio seems to support?

I found this MBC3 one from IG, but it's discontinued and was was not sold pre-assembled ("Build it yourself"). It looks like it may be the one GBStudio supports based on the enable bit value.

> How to use the rumble motor
> You need to write the byte 0x20 to 0x4000-0x5FFF. Usually games write 0x08 but that would conflict with the RTC so you need to patch games to use 0x20 instead. For [Pokemon Pinball we have an example patch](https://shop.insidegadgets.com/wp-content/uploads/2020/04/Pokemon-Pinball-UE-Rumble-0x20-Patch.zip) to apply but it doesn’t appear to activate the rumble strong enough – you could lower the resistor if need be.

https://shop.insidegadgets.com/product/gameboy-4mb-32kb-fram-mbc3-with-rtcrumble-plus-2x-2mb-mode-flash-cart-build-it-yourself/

If that is the case then the unofficial MBC3 rumble variant may be obscure enough that it doesn't merit support in the example.

EV3lindaboi:
Remade the whole example. How is it now?

untoxa:
sorry, what? https://github.com/gbdk-2020/gbdk-2020/pull/789/commits/4531d5eb2b14d49a230022b4b49ffa25235a57cb#diff-5466e8b68e0e5a551bb6c430dd62d228f794a5ce593518a59b691056d6bc70c7R3-R9

EV3lindaboi:
crude attempt of changing the intensity of the rumble

bbbbbr:
We'd like to see a step above a crude attempt to merge an example. :)

The option for rumble intensity should be controlled not by a blocking delay, but varying the duty cycle on a per-frame basis which gets updated once per frame along the lines of my example above. 
  - And with constants defined for the intensity levels, something like `RUMBLE_LOW`, `RUMBLE_MED`, `RUMBLE_MAX`

The readme could have a list of MBC's which support Rumble and where to edit them in the `Makefile`
```
// Rumble requires MBC set to one of the following:
// 0x1C   MBC-5                    RUMBLE   8 MB
// 0x1D   MBC-5   SRAM             RUMBLE   8 MB
// 0x1E   MBC-5   SRAM   BATTERY   RUMBLE   8 MB
```

There could be ON/OFF MACROS and constants for the enable bit
```C
#define MBC_RUMBLE_BIT_ON   0b00001000u
#define MBC_RUMBLE_BIT_OFF  0b00000000u
#define MBC5_RUMBLE_ON      SWITCH_RAM_MBC5_LOCAL(MBC_RAM_BANK_0 | MBC_RUMBLE_BIT_ON)
#define MBC5_RUMBLE_OFF     SWITCH_RAM_MBC5_LOCAL(MBC_RAM_BANK_0 | MBC_RUMBLE_BIT_OFF)
```

As mentioned earlier, for some of this I'm looking at my implementation in canyon racer as a reference:
https://github.com/bbbbbr/canyon-racer/blob/main/src/mbc5_rumble.h
https://github.com/bbbbbr/canyon-racer/blob/1372db39392f5635ed773b02ab0dea7938f6d29a/src/audio/audio.c#L130C1-L142C11

bbbbbr:
OK to merge now with the understanding that I'll make changes afterward that are easier to do directly.

PR #850, opened 2025-12-20
Remove duplicated line from text_advanced_dialogue example

michel-iwaniec:


bbbbbr:
Yep, confirming it is duplicated in line 159

PR #856, opened 2026-01-17
Optimisation of divisions

Phidias618:
I optimised all of the division functions by using a faster algorithm for unsigned division.
I also fixed a bug involving the carry flag not always being set when doing a signed division by 0.

The functions .div8, .mod8, .div16, .mod16 still exists in order to maintain compatibility, 
but they are now wrappers arround the newer functions.

### DIFF
--- a/gbdk-lib/libc/asm/sm83/div.s
+++ b/gbdk-lib/libc/asm/sm83/div.s
@@ -67,8 +67,8 @@ __moduint:
 
 .divmod_uchar_bcde:
 __divuchar:
-        ld		c, e
-        ld      e, a
+        ld	c, e
+        ld  e, a
         xor a
         ld b, a
         ld d, a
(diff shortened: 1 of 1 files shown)


bbbbbr:
(Duplicating this from Discord)

Added the output from the PR version to the comparison spreadsheet (which is now too big to upload but I've attached the debug log output for both).

Here's a summary of the output, with the caveat that the test program shared earlier could use review by others.

- (u8 ,u8) looks good: Old and PR match, and they align with spreadsheet version
- (u8, s8) Old and PR have same output, but both vary some vs spreadsheet: 255 for div, 31310 for mod
- (s8, u8) and (s8,s8) Old and PR have a lot of differences in div and mod, a lot of the divs look off in the PR version (sometimes just sign, other times the value)

[divmod_test.zip](https://github.com/user-attachments/files/24694939/divmod_test.zip)

divmod_pr_s8_u8_compare
divmod_pr_s8_s8_compare

bbbbbr:
Yeah, looks good for 8 bit now (with the 6th commit).

16 bit is looking good too. I did ~7.2 million samples and the PR v6 output looks identical to the old GBDK implementation (while being faster).

PR #859, opened 2026-01-28
Optimization of multiplications

Phidias618:
There are 2 new optimizations to __mulint
- All of the loops are now unrolled
- A multiplication is now computed in a different but faster way

Previously a 16 bit multiplication required about 200 M-cycles, now it requires less than 120 M-cycles.
about 2/3 of this improvement comes from the unrolling of the loops, the other 1/3 comes from the algorithm change.

### DIFF
--- a/gbdk-lib/libc/asm/sm83/mul.s
+++ b/gbdk-lib/libc/asm/sm83/mul.s
@@ -29,7 +29,8 @@
 ;--------------------------------------------------------------------------
 
         ;; Originally from GBDK by Pascal Felber.
-
+        ;; Updated by Phidias618.
+        
         .module mul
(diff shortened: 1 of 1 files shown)


bbbbbr:
Thanks for the PR! Will try to do some testing/validation on this soon-ish

bbbbbr:
8 Bit output: looks good!
- ~17% to 23% less cycles

16 Bit output: looks good!
- ~26% to 30% less cycles

ROM0 size
- 62 bytes larger

I'd recommend restoring (updating if needed) the entry/exit comments on the various labels for `__mulint:`, `__muluchar`, `etc`. They're helpful when coming to the source without previous familiarity

PR #861, opened 2026-02-15
Bug fix + Further optimization of divisions (testing required)

Phidias618:
The worst case for this code runs in about 280 M-cycles, as opposed to a worst case that runs in about 580 M-cycles for the previous division code.
It also fixes a bug where a division by one could make the program run endlessly if the dividend is large enough.

### DIFF
--- a/gbdk-lib/libc/asm/sm83/div.s
+++ b/gbdk-lib/libc/asm/sm83/div.s
@@ -49,89 +49,104 @@ __divuchar:
         ld d, a
         ; Fall through .divmod_uint_bcde
 .divmod_uint_bcde:
-
-
-	    ; stores -Y in hl and check that -Y is not zero
+__divuint::
+		; computes the quotient and the remainder of X / Y
+        ; X is stored in de
+        ; Y is stored in bc
+        ; outputs the quotient in bc
+        ; outputs the remainder in de
+        ; if Y = 0 then the carry is set and quotient = 0 and remainder = 0
+        ; otherwise the carry is cleared
+		
+	    ; stores -Y in hl
         xor a
 		sub c
 		ld l, a
 		sbc a
 		sub b
 		ld h, a
+
+		inc a
+		jr z, small_y				; jump if 1 <= Y <= 256
 		
+		ld a, h
 		or l
-        jr z, .division_by_zero
+		jr z, division_by_zero
 
-
-		inc h
-		jr z, 0$
-		dec h
+		; if this part of the code is reached then Y > 256
+		; Thus the quotient will fit within 8 bits
 		xor a
-		;ld b, a
-4$:
+		ld b, a				; set the upper byte of the quotient to 0
+
+		; computes the largest K such that ((-Y) << K) fits within 16 bits
+		; ((-Y) << (K + 1)) will be stored in hl
+		; K will be placed as a base 1 number using the most significant bits of C
+0$:
 		rra
 		add hl, hl
-		jr c, 4$
-		;ld c, a
+		jr c, 0$
+		ld c, a
 
-		;;ld bc, hl
-		;;ld hl, de
-
-		;;ld e, a
-		;;ld d, #0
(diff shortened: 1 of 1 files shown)


bbbbbr:
For the issue mentioned below, were there specific values that reproduce the issue?

> It also fixes a bug where a division by one could make the program run endlessly if the dividend is large enough.

Phidias618:
> For the issue mentioned below, were there specific values that reproduce the issue?
> 
> > It also fixes a bug where a division by one could make the program run endlessly if the dividend is large enough.

computing 65535 / 1 should be the only division that runs endlessly, and any division by 1 with a quotient greater or equal than 32768 might return wrong results

bbbbbr:
Tested 8 and 16 bit output and those look good.

Benchmarking the full 8 bit ranges against `develop` shows what looks like a significant regression in performance, taking about 60% more cycles.

```
develop
_test_divmod_u8_s8      103,056,536
_test_divmod_s8_s8       91,882,460
_test_divmod_s8_u8       81,715,956
_test_divmod_u8_u8       63,725,236

patch-5
_test_divmod_u8_s8      167,804,056
_test_divmod_s8_s8      169,922,308
_test_divmod_s8_u8      160,969,716
_test_divmod_u8_u8      133,908,148
```
I'll try to benchmark a subset of the 16 bit range next.

bbbbbr:
Benchmarked a small subset of the 16 bit range. In aggregate some of it looks like a wash (~2-3%), some looks ~12% more cycles. So a smaller regression than 8 bit, at least for the values covered by the testing.

```
develop
_test_divmod_s16_s16        2,508,252,864
_test_divmod_s16_u16        2,453,049,020
_test_divmod_u16_s16        1,917,571,468
_test_divmod_u16_u16        1,871,568,264

patch-5 16 bit
_test_divmod_s16_s16        2,583,366,792
_test_divmod_s16_u16        2,528,162,948
_test_divmod_u16_s16        2,155,741,772
_test_divmod_u16_u16        2,109,738,568
```

bbbbbr:
Given the benchmarks above can the bug fix be split out from the other changes? We'll definitely want the issues fixed but I'm not as sure about the other optimizations.

untoxa:
Yes, i agree. @Phidias618 Can we fix the issues with the current division algo?

Phidias618:
> Given the benchmarks above can the bug fix be split out from the other changes? We'll definitely want the issues fixed but I'm not as sure about the other optimizations.

Thank you for comparing the speed of the 2 algos. I was so focused on improving the worst case of the algorithm that i worsened the average case.
Before i revert these so called "optimizations" there is one last thing that might improve performances that i want to try.
If this still isn't an improvement then i will just use the previous algo with an additional check for a divisor of 1

Phidias618:
> Yes, i agree. @Phidias618 Can we fix the issues with the current division algo?

Yes, this is just a matter of checking whether the divisor is equal to 1 inside the division routine

untoxa:
please make the separate pull request which fixes the known issues, then we'll play with optimizations, because the current experimental build is already in use by the third party (gbstudio experimental and crosszgb experimental)

Phidias618:
> please make the separate pull request which fixes the known issues, then we'll play with optimizations, because the current experimental build is already in use by the third party (gbstudio experimental and crosszgb experimental)

I made a separate PR that just fixes divisions by 1, the fix is not pretty but it should work

bbbbbr:
Now that the fix has been split and merged in #865 lets close this one and if there is another pass on the optimizations it can be in a fresh PR rebased on the changes in `develop`.

If a new PR is opened it should include the following:
- Confirmed passing the entire ranges of u8/u8, u8/s8, s8/u8, s8/s8 tests
- A reasonable amount of confirmed passing test coverage for 16 bit values (can be the small test app I shared earlier or something else)
- Some benchmarks for the above showing the aggregate performance differences

PR #862, opened 2026-02-20
improve png2asset tile palette handling

untoxa:


bbbbbr:
Noting a couple things about this change for future reference:

The functional part of the fix appears to be located in this one changed line:
https://github.com/gbdk-2020/gbdk-2020/pull/862/changes#diff-d4189ea1c30486e6a0434b602e8e6c16c3924823c1c35638e01f391d659bd5da
```diff
-        return data == t.data && pal == t.pal;
+//        return data == t.data && pal == t.pal; // probably, sometimes we need to take palette into account?
+        return data == t.data;
```

The `zero_palette` flag appeared to be related to `pal_idx` (palette number) getting re-calculated for Maps with attributes and Metasprites in the two locations below. 

In: GetMetaSprite()
https://github.com/gbdk-2020/gbdk-2020/blob/b83431218fa0006f01cb0a84f6cdc2d30b635875/gbdk-support/png2asset/metasprites.cpp#L40

In: GetMap()
https://github.com/gbdk-2020/gbdk-2020/blob/b83431218fa0006f01cb0a84f6cdc2d30b635875/gbdk-support/png2asset/maps.cpp#L123C12-L125C175
```C
 if(assetData->args->use_map_attributes)
            {
                unsigned char pal_idx = assetData->image.data[y * assetData->image.w + x] >> assetData->args->bpp; //We can pick the palette from the first pixel of this tile
``` 
The palette is now calculated in the tile itself and retained there instead of zeroed on some occasions, with the above code changed to use that value.

The palette for a tile is now picked based on the Maximum palette found in a tile instead of the upper-left.

PR #871, opened 2026-03-15
GB: Drawing.s: box: fix drawing top and bottom incorrectly when x1==x2

bbbbbr:
- Output when x1==x2 was an I beam shape instead of a 1 pixel wide line

### DIFF
--- a/gbdk-lib/libc/targets/sm83/drawing.s
+++ b/gbdk-lib/libc/targets/sm83/drawing.s
@@ -539,7 +539,7 @@ ycirc$:
         LD      A,(.x_s+1)
         LD      C,A
         SUB     B
-        JR      NC,ychk$
+        JR      NC,ychk$   ; If x1 > x2 then swap
         LD      A,C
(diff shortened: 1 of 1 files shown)


gbdk-lib/libc/targets/sm83/drawing.s
```diff
@@ -563,56 +563,64 @@ dbox$:
...
+        ; Shrink top and bottom of box lines inward by 1 to not draw those pixels twice
+        ; Skip if x1 == x2 (width 1) since it would erroneously widen (instead of shorten) them
+        ; Width 2 is ok since it ends up swapping x1 and x2, which draws ok
+        LD      A,(.x_s+1)
+        LD      B,A
         LD      A,(.x_s)
-        INC     A
+        CP      B
+        JR      Z,1$
+        INC     A          ; x1 += 1
```
bbbbbr, line 583: This and the few lines above are the only functional change, the rest are just adding comments

PR #872, opened 2026-03-22
Optimization of uitoa and of itoa for the sm83 port

Phidias618:
.uitoa is now much faster, here is an estimation of the number of required M-cycle to convert an n-digit number to ascii
- 1 digit  : 40ish cycles
- 2 digits : 140ish cycles
- 3 digits : 230ish cycles
- 4 digits : 340ish cycles
- 5 digits : 430ish cycles

In comparison the previous implementation of .uitoa required at least 500 M-cycles to finish, regardless of the size of the input number.
The new implementation is also 4 bytes smaller.
uitoa and itoa now uses the  __sdcccall(1) calling convention.

### DIFF
--- a/gbdk-lib/include/asm/mos6502/stdlib.h
+++ b/gbdk-lib/include/asm/mos6502/stdlib.h
@@ -1,0 +1,151 @@
+/** file stdlib.h
+    'Standard library' functions, for whatever that means.
+*/
+#ifndef STDLIB_INCLUDE
+#define STDLIB_INCLUDE
+
+#include <types.h>
+
+/** Causes normal program termination and the value of status is
+    returned to the parent.
+    All open streams are flushed and closed.
+*/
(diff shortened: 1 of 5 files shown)


untoxa:
if you split the stdlib.h across the ports, then you should clean those up, because OLDCALL has no meaning on the 6502 target, for example. not just copy. that may be merged, but further work is required.

untoxa:
REENTRANT make no sense for the Z80 derivative targets...

Phidias618:
> if you split the stdlib.h across the ports, then you should clean those up, because OLDCALL has no meaning on the 6502 target, for example. not just copy. that may be merged, but further work is required.

I noticed that string.h for the 6502 port also contains reference to OLDCALL, i didn't remove it in this PR because i don't think this belong inside this PR

PR #877, opened 2026-05-02
Halved size of delay.s for the sm83 port

Phidias618:
The new code is more than half as small (20 bytes instead of 45 bytes)
The new code uses less stack, the maximum stack depth is smaller by 2 bytes.
The previous algorithm rounded a millisecond to 1048 cycles, the new algorithm rounds one millisecond to 1048.5 cycles whereas the true duration of a millisecond is 1048.576 cycles. As such the new algorithm is slightly more accurate

### DIFF
--- a/gbdk-lib/libc/targets/sm83/delay.s
+++ b/gbdk-lib/libc/targets/sm83/delay.s
@@ -8,61 +8,29 @@
 	;;   DE = number of milliseconds to delay (1 to 65536, 0 = 65536)
 	;; 
 	;; Register used: AF, DE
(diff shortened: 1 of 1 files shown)


untoxa:
Why not making it producing the same delay in the double sped mode instead?

Phidias618:
> Why not making it producing the same delay in the double sped mode instead?

That's now done. I also removed 0.5 cycle per loop iteration as i am not sure whether this kind of precision will ever be required.
The new code is 25 bytes long, which is still almost twice as small as the old code.

PR #878, opened 2026-05-02
Updated the calling convention of memset and small optimization of memset for the sm83 port

Phidias618:
I updated the calling convention of memset from __sdcccall(0) to __sdcccall(1)
I also optimized a tiny bit the function itself, it is now a few bytes smaller and it is faster by a few cycles.
I removed the code that checked for a length of 0 because it was not needed, the code without that check already handle well a length of 0.


untoxa:
i think this should not work in the n==0 case, also i have doubts about the other cases. need tests.
edit: or, i might be wrong, but good to have unit test anyway.

PR #882, opened 2026-05-07
Optimization of reverse.s for the sm83 port

Phidias618:
The new function is faster for every string length, it is 13 bytes smaller and it don't use the stack at all.
It also uses the calling convention __sdcccall(1) instead of OLDCALL


untoxa:
i just noticed - we are missing reverse() function implementation for the Z80 target, will be nice to add that one as well

Phidias618:
> i just noticed - we are missing reverse() function implementation for the Z80 target, will be nice to add that one as well

That explains why i wasn't able to find the implementation of reverse for the z80 target

Phidias618:
> i just noticed - we are missing reverse() function implementation for the Z80 target, will be nice to add that one as well

I finished implementing reverse for the z80.
I haven't tested the program, but i tested the sm83 version of it, which is mostly the same except for a loop which has been replaced by a cpir.

Phidias618:
I realized i forgot to add reverse.s to the Makefile for the z80

PR #883, opened 2026-05-07
Small optimization of rle decompress functionn

Phidias618:
I replaced some instructions with faster one. The new code is as such a tiny bit faster and smaller.
(i already made a similar pull request a few days ago, but i only realized now that i could implement the exact same optimization for these files as well)


untoxa:
this function is never used in the modern SDCC - generation of compressed initializers is disabled in the SDCC sources

PR #884, opened 2026-05-08
Small optimization of memcpy for the z80 port

Phidias618:
The new function is 24 cycles faster.
If the length argument is a multiple of 16 then the function is faster by an additional 10 cycles.
The new function is also 6 bytes smaller


untoxa:
early exit must crash because of the wrong value on the top of the stack

PR #905, opened 2026-06-13
NES mapper configuration support

michel-iwaniec:
- Define new mapper targets:
  - nes-mapper30-s-lomem (mapper30 single-screen layout, LOMEM config- current target)
  - nes-mapper30-s-win (mapper30 single-screen layout with window emulation)
  - nes-mapper30-f (mapper30 four-screen layout)
  - nes-mapper2-h / nes-mapper2-v (mapper2 horizontal / vertical layout)

- Support "nes" an an alias of nes-mapper30-s-lomem for backwards compatibility

- Copy everything but platform_cfg.s from gbdk-lib/libc/targets/mos6502/nes/ to top-level directory gbdk-lib/libc/targets/mos6502/
- Create new directories for the new targets gbdk-lib/libc/targets/mos6502/nes/mapper30-s-win gbdk-lib/libc/targets/mos6502/nes/mapper2-h

- Update lcc/targets.c with new configurations

- Update Makefiles

- Update docs to describe supported mappers and trade-offs

- Update gbdk-lib/examples/cross-platform/large_map/ with new targets to illustrate scrolling glitches


bbbbbr:
Just from looking at the changes but not testing them locally: If the Makefile is changing the obj and bin subdirs from `EXT` to `PLAT`, then on the `<target>-clean` you probably(?) also need to pass/define  `PLAT` so it can find the right subdirs to remove.

michel-iwaniec:
> Just from looking at the changes but not testing them locally: If the Makefile is changing the obj and bin subdirs from `EXT` to `PLAT`, then on the `<target>-clean` you probably(?) also need to pass/define `PLAT` so it can find the right subdirs to remove.

Ah, good catch! Should be fixed now.

bbbbbr:
Looks good to me :+1:

Issue #911, opened 2026-08-19
Compile fails with security error on cc1.


devnevins:
When compiling under MacOS (ARM) there is an error in the compilation process. The error "... developer cannot be verified, macOS cannot verify that this app is free from malware" appears for the program cc1. This program is not under the bin directory; it's under the libexec/sdcc directory. In order for a sucessful compilation the xattr com.apple.quarantine must be removed from cc1. That step needs to be added to the documentation.

bbbbbr:
Thanks. Added one suggestions on it, but otherwise it looks good.

bbbbbr:
Hmm actually not sure that code review I added to it would work. I've deleted it. 

Just change this part `bin and sdcc folders`  to `bin and libexec/sdcc folders, and it should be good to go.

PR #912, opened 2026-08-19
Fixed cc1 xattr for MacOS compiles.

devnevins:
This fixes issue #911 by adding a little text to the documentation detailing the xattr removal for the cc1 program.


bbbbbr:
Thanks. One suggestion. 

Change this part `bin and sdcc folders` to `bin and libexec/sdcc folders`, and it should be good to go.

devnevins:
That's a good suggestion. I've implemented it and pushed the change.

bbbbbr:
Thanks!

PR #913, opened 2026-08-21
Improvement of integer to ascii conversion for the sm83

Phidias618:
Here are the speed of the previous implementation of utoa
- 1 digit : 40ish cycles
- 2 digits : 140ish cycles
- 3 digits : 230ish cycles
- 4 digits : 340ish cycles
- 5 digits : 430ish cycles

Here are the speed of the new implementation in the average/worst case
- 1 digit : 51/51 cycles
- 2 digits : 80ish/103 cycles
- 3 digits : 110ish/162 cycles
- 4 digits : 150ish/227 cycles
- 5 digits : 190ish/276 cycles

Except for the 1 digit case, which is a bit slower, every other case is much faster.
The new implementation is also 22 bytes smaller.

### DIFF
--- a/gbdk-lib/libc/asm/sm83/itoa.s
+++ b/gbdk-lib/libc/asm/sm83/itoa.s
@@ -40,135 +40,89 @@ _uitoa::
     push hl               ; put the return address back
 .utoa::                         ; convert unsigned int into ascii
     ; input :
-	;  - de (x) : the 16 bit unsigned number to convert to ascii
-	;  - bc (dst) : the pointer to the destination string
-
-    ld h, b
-	ld l, c
-
-	push bc
-	ld bc, #10
-	jr 4$
-0$:
-	push hl
+	;  - de : the 16 bit unsigned number to convert to ascii
+	;  - bc (dst) : the pointer of the destination string
+	; bc is preserved by this function
 	
-	; compute an approximation of x / 10
-	; by using the approximation 1/10 ~ 0.0001100110011
-	
-	ld b, e
-	ld c, d
+	ld a, #'0'
+	push bc				; save dst
 
-	ld l, e
-	ld h, d
-	xor a
-.rept 3
-	add hl, hl
-	adc a
-.endm
-	ld e, h
-	ld d, a	
-	
-	add hl, hl
-	adc a
-	ld l, h
-	ld h, a
+	ld hl, #-10000		
 	add hl, de
+	jr c, 0$			; 5 digit number ?
 	
-	ld e, b
-	xor a
-	ld b, a
+	ld hl, #-1000
+	add hl, de
+	jr c, 2$			; 4 digit number ?
 	
-	add hl, bc
-	srl c
-	add hl, bc
-
-	srl c
-	srl c
-	srl c
-	add hl, bc
-	srl c
-	add hl, bc
+	ld hl, #-100
+	add hl, de
+	jr c, 4$			; 3 digit number ?
 	
-	; computes an approximation of the remainder 
-	; by using the computed approximation of x / 10
-	; it will always be in the range [0; 75]
-	sub l
-	add a
-	add a
-	sub l
-	add a
-	add e				; e = x - 10 * hl
-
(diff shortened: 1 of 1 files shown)


bbbbbr:
Given that there have been bugs found and fixed as part of the process for some of the previous PRs we'd like to see some kind of extensive test coverage (as an attachment to the ticket) that helps validate correctness for each of these library PRs.

It's expected of course that the difficulty of setting up tests may vary based on the particular code and whether it's interacting with things like safe vram timing, etc.

That helps lower the work somewhat on our end. We appreciate the contributions to optimizing the library, and it has to be balanced against how much time and effort is required on our end trying to ensure that merged PRs don't introduce regressions.

Phidias618:
> Given that there have been bugs found and fixed as part of the process for some of the previous PRs we'd like to see some kind of extensive test coverage (as an attachment to the ticket) that helps validate correctness for each of these library PRs.
> 
> It's expected of course that the difficulty of setting up tests may vary based on the particular code and whether it's interacting with things like safe vram timing, etc.
> 
> That helps lower the work somewhat on our end. We appreciate the contributions to optimizing the library, and it has to be balanced against how much time and effort is required on our end trying to ensure that merged PRs don't introduce regressions.

Could i know what went wrong with the other PR such that it got reverted?

As for this PR, i used the exact same tests as my previous PR that changes integer to ascii conversion to verify the correctness of the new algorithm. I might as well share it again here (i already shared it on discord a few month ago, but its easier that i just share it again here)
[new_uitoa_test.txt](https://github.com/user-attachments/files/31355541/new_uitoa_test.txt)
I can confirm that all test passes.
As for the cycle count i gave, i apologize because the new counts are a bit off, but not by much. 
It turns out i overestimated a few duration (but none has been underestimated though, which means this PR is even better than i expected). 
Each average duration is accurate with a margin of 5 cycles, except for the last one, which is closer to 180 cycles than to 190.
As for the maximum durations, here is the new table
- 1 digit : 51 cycles (that has not changed)
- 2 digits : 96 cycles
- 3 digits : 161 cycles 
- 4 digits :  227 cycles (that has also not changed)
- 5 digits : 269 cycles
The new cycle figures were obtained by simply using an emulator that counts cycles and by running it on every possible input, instead of counting cycles by hand (which is inefficient and error prone). I should have done that from the begining, this would have saved me a lot of time.
As a quick sanity check i also checked the accuracy of the speed of the previous implementation, its accurate up to a margin of 10 cycles.

bbbbbr:
With regard to the reverted PR, @untoxa has the details. It sounds like it broke something in CrossZGB.

untoxa:
commit was reverted, because wrapping did not work correctly. CrossZGB backgrounds in games became messed up, i did not have time to investigate and fix, so i had to revert.

Phidias618:
> commit was reverted, because wrapping did not work correctly. CrossZGB backgrounds in games became messed up, i did not have time to investigate and fix, so i had to revert.

Thanks for pointing that out. 
Does the bug still happen when Y coordinate specified in .set_xy_tt is in the [0, 31] range? I am asking because i assumed the Y coordinate to always be in that range, and if it's not then the behaviour is not the same as the previous implementation.
If its not that, then which wrapp arround is not working? The vertical one? the horizontal one? both? Could you share a screenshot?

bbbbbr:
> I am asking because i assumed the Y coordinate to always be in that range, and if it's not then the behaviour is not the same as the previous implementation.

This prompts a couple more things I'd like to see for these PRs going forward:
- What assumptions are made about the inputs, and whether those differ from the previous implementation (or at least what they are understood to be for the previous)
- **Any** changes to output behavior, whether intended or (what is thought to be) potentially unintended

bbbbbr:
> As for this PR, i used the exact same tests as my previous PR that changes integer to ascii conversion to verify the correctness of the new algorithm. I might as well share it again here (i already shared it on discord a few month ago, but its easier that i just share it again here)

Yes, please always attach the materials to the PRs and mention how the testing went. It's much easier to follow that way since things get easily buried in discord.

PR #918, opened 2026-08-24
Fix typo loose to lose

systemoflevers:
Fix common typo of loose instead of lose.


bbbbbr:
Thanks!

While you're at it, I see another typo in the area your correction covers. Could you also delete the word `remove` in this?
> `you could remove adjust your tileset`

systemoflevers:
np, done

PR #921, opened 2026-09-19
Fix bug with non-clean build introduced by PR#905

michel-iwaniec:


bbbbbr:
Fix looks good, merging

